PR #9, opened 2018-08-14
expand hat guessing

fpvandoorn:


fpvandoorn:
the current discussions on Discord are not included yet.

PR #10, opened 2018-08-20
Add note to Baton pass about setting up finesses.

alercah:


Zamiell:
This doesn't apply. You can't perform a layered finesse if a card inside the later is unplayable.

alercah:
But it might become playable as part of the finesse resolving.

Zamiell:
While possible, I've never seen that occur. The document, while intended to be thorough, is not intended to document every edge case. So it probably doesn't need to include that as a possible reason to baton discard.

PR #25, opened 2019-04-18
Typo: Suit-change in middle of example

hunt0r:
Accidentally changed the "target suit" from red to green in middle of an example.


Zamiell:
i'll fix it manually, thanks

PR #41, opened 2019-09-23
Cleanup of stalling-related conventions

alercah:
- Clarify the list of allowable *Tempo Clues* by grouping the list of allowable stall situations together, as well as merging the two entries that relate to unlocking locked hands.
- Add a clue that gets at least one untouched card played to the list of allowable *Tempo Clues* explicitly; it follows from *Minimum Clue Value Principle* 
- Add a section on 7 Clue Tempo Clues. This is intended only as an elaboration on existing conventions, not anything actually new.
- Rearrange the sections following the *Tempo Clues* section so that they appear in order of interpretation priority, with stalls first, then *Double Tempo Clues*, then *Tempo Clue Trash Moves*
- Clarify and simplify discussion of relative interpretation priority of *Tempo Clues*, *Double Tempo Clues*, and *Tempo Clue Trash Moves* in their respective sections, with some minor formatting and grammar fixes.
- Add a "7 Clues" as severity 2 stalls to the stall table, bumping up other severities. Only *Tempo Clues* are allowed at 7 clues.
- Rename the "Double Discard" column on the stall table to "Unsafe Discard", because it also applies to *Scream Discard* situations. To ensure there's no risk of this being applied incorrectly, add text clarifying that a player cannot stall on a hunch that they might have a critical card on chop; there must be a convention blocking them from discarding.
- Clarify some points on the stall table: the *Finesse Position Exception* always applies to *5 Stalls*, regardless of reason, and stalls are not allowed to be given when a player can play.
- Add discussion and examples of the implications of the *Schrödinger's Cat Principle* on stall situations, since they are far and away where the it causes the most confusion and misplays.
- Add explicit text that breaking the stall priorities is not allowed. Still subject to the general "every rule can be broken" rule, but stall priority is about as close as we get to it being an absolute rule.
- Clarify that a hand isn't locked if there is a known play.
- Rearrange the *Locked Hand Saves* section to be more clear.
- Elaborate and clarify a little on the effect of a *Locked Hand* on preceding players, including that *Anxiety Play* is the only exception to leaving them with at least one clue, and the effects of the *Guide Principle*.
- Also mention locked hand unlocking with *Tempo Clues* in the *Locked Hand Saves* section since it's the primary discussion location for *Locked Hands*.

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -398,48 +398,49 @@ Note that in this user interface, players are holding their hands like they woul
 ### Tempo Clues (Re-cluing a Card)
 
 * *Tempo Clues* are defined as clues that touch no new cards. Their purpose is to get 1 or more cards played that already had a clue.
-* Note that *Tempo Clues* do not meet *Minimum Clue Value Principle*. They are inefficient, and in Hanabi you want to be as efficient as possible. Thus, we agree that *Tempo Clues* are only allowed in the following special circumstances:
-  1. When the tempo clue gets 2 or more clued (or unclued) cards played
-  2. When the card is "out of order" (meaning that it is impossible for a *Prompt* to get the cards played)
-  3. When the card was *Chop Moved* and it is now playable
-  4. When the clue giver has a fully clued hand
-  5. When the clue giver has been *Screamed* at
-  6. When the clue giver is in a *Double Discard* situation (see *Avoiding Double Discards* below)
-  7. When the clue receiver has a fully clued hand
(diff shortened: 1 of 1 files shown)


Zamiell:
Can you break this up into a bunch of smaller prs? It is much easier to work with

Zamiell:
like one pr per section or something

Zamiell:
i roughly integrated this PR into the main document, submit PRs with further changes if you wish

PR #169, opened 2020-03-24
Remove copy+paste error in text.

EhDrone:
Removing text "breaking clues number 5 to invarse. That touches the blue 5. Since the blue 5 is one-away from chop, invarse will think that this is a *5's Chop Move*." that seems to be a copy+paste error.


EhDrone:
Already opened as a pull request - apologies for duplication.

PR #172, opened 2020-03-31
fix name in Clandestine Finesse example

cg505:
This looks like a mistake left over from 54e3125093072ee089124d41f25e829edfb1d121.


Zamiell:
thanks

PR #258, opened 2020-06-03
Update Null.md

TheDaniMan:
Normal colours are always ordered red, yellow, green, blue, purple, if I'm not mistaken, but this example suggests otherwise. Also, is it weird to have the exact same example twice?


Zamiell:
thanks

PR #265, opened 2020-06-05
Update Reference.md

TheDaniMan:
UTD should be read before this (though I'm not sure if I did the link right). Why is UTD below this part anyway?


Zamiell:
level 15 isn't meant to be read in order

PR #272, opened 2020-06-06
Clarify LHS stall priority

padiwik:


Zamiell:
you added a bullet point mentioning LHS, but LHS is not defined yet at that point in the document

padiwik:
that complaint doesn't make sense when it's present in the table above

padiwik:
Maybe a better solution is to define LHS, 8 Clue Save, and Filling in info as having the same precedence? Since technically you can decline to perform an 8 clue save and fill in info instead

Zamiell:
> Maybe a better solution is to define LHS, 8 Clue Save, and Filling in info as having the same precedence?

that kind of deletes all of the columns then

Zamiell:
ok i updated the doc, let me know if this good

padiwik:
seems ok, I'd appreciate having the note on why they're tied but it's not strictly necessary.

actually now I'm thinking about the edge case where a lhs is available but no fill ins. Alice does a hard burn because she doesn't want to LHS, so technically the priority table is wrong. but it's good enough

PR #292, opened 2020-06-18
fix apparent bad touch in prompt2

xcpenguin:
Prompt 2 example has apparent bad touch, which would confuse beginners even if it's valid for advanced paly. I changed the third red clue to unknown with negative red clue.


Zamiell:
thanks

PR #293, opened 2020-06-21
Fix two outdated references

eljobe:
In the last bullet of the "The Early Game (Intermediate Section)" there was a parenthetical note about "Order Chop Move" being an "Advanced" move that would be introduced later. But, it's actually already been introduced in Level 5 - Apprentice.

Also, below the "Allowable Stall Clues" table, there was a reference to Hard Burns being precedence 7. When, the table actually shows them as precedence 5. (I assume they used to be precedence 7, but that the reference wasn't updated when precedences 4-6 were collapsed into a single tied precedence 4.


Zamiell:
perfect, thank you very much

PR #295, opened 2020-06-22
Corrections for "The Ambiguous Finesse Pass-Back"

eljobe:
Here, I correct one type adding the word "after" to the phrase "after all".

I've also taken the liberty to directly link to "Layered Finesse" and to name the level in which the convention is actually introduced. In the old version, it says that it is covered "in a later section.", but it's actually covered earlier in the document.


Zamiell:
looks good, thank you

PR #306, opened 2020-07-03
Add link to explain play order inversion

padiwik:
It's a common question, and I don't think it's harmful to have it in the main reference document


Zamiell:
add the other ones then?

padiwik:
meh there isn't really a good place for chop focus, and 3 bluffs is more history than logic. my thinking was anyone who cares would have seen this link

PR #308, opened 2020-07-03
no play clues on 5s

TheDaniMan:
a. updated anything that mentions play clues on 5s in the early game due to them having been made illegal. You should probably doublecheck the changes I made to 5nd.
b. grammar stuff in top hat clue

There were also two references to play clues on 5s prior to level 6. However, the "no play clues on 5s" section is in level 6, so I did not touch them.


Zamiell:
> There were also two references to play clues on 5s prior to level 6.

where

TheDaniMan:
Oh I was looking under "The 5's Chop Move" and "Double Discard Situations", but I just realized neither of these necessarily pertain to early game so they're fine anyways. I double-checked the changes I actually made though, and I think those are fine.

Zamiell:
thank you!!!

PR #310, opened 2020-07-11
Fix broken links in learning path.md

squidgetx:
Since the learning path document got moved all the relative links got broken.


Zamiell:
thanks

PR #314, opened 2020-07-14
Fix Hat Guessing example

padiwik:


padiwik:
oh i see why it was bugged https://github.com/Zamiell/hanabi-conventions/commit/ca2051c2b7ee03df7a6d6c6c3526f1424c38179f

PR #351, opened 2020-08-30
Grammar: change amount to number

padiwik:
This has bothered me for a while

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -2106,7 +2106,7 @@ There are a lot of different moves that involve cluing trash. [Here is a handy f
 * With that said, it is sometimes possible to construct **safe** lines that contain a *Lie Component*. There is usually deep levels of nuance involved in these lines, so this kind of thing is only recommended for advanced and expert players.
 * The previously mentioned *Out-of-Order Finesse* is a specific example of a *Finesse* with a *Lie Component*. One key attribute of the *Out-of-Order Finesse* is that the player who receives the *Fix Clue* knows that **all** of the related cards are playable, or it would not have been worth the risk (and the confusion) to perform a line with a *Lie Component* in the first place.
 * This concept can be generalized to all *Finesses* that have a *Lie Component*. After a player receives a *Fix Clue* in such a situation, they should assume that the *Finesse* is "still on". So, depending on the situation, they should either know that the initial card that was clued is playable (like in an *Out-of-Order Finesse*) or continue to blind-play cards to fulfill the initial *Finesse* (if the card clued with a *Fix Clue* seems to be unrelated).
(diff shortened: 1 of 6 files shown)


Zamiell:
its clearer the way it is because "number" refers to a "number clue"

padiwik:
It is clear that "number of clues" and "number clue" are distinct. It also sounds wrong when amount is used here (https://grammarist.com/usage/amount-number/)

padiwik:
(and aside from the "different number of positive/negative clues" in the Order Chop Move section, there is no ambiguity from saying "number of cards", "number of players", etc)

Zamiell:
```
* Different numbers of negative clues are ignored.
```
this is really confusing. it looks like it is talking about "negative 2", "negative 3", etc

padiwik:
hmm i see.. can it just be "different negative clues" and still have the same meaning?

Zamiell:
im not going to sacrifice clarity for grammar

padiwik:
specifically for that OCM section: even as written it is a mouthful, so the phrasing should be simplified if possible

Zamiell:
i dont think "different negative clues" is very clear

padiwik:
currently ("different amount of negative clues"):
look at all the negative clues, count them, then see if the counts match

proposal ("different negative clues"):
look at all the negative clues and see if there are any differences

is the second not simpler? at least i don't see how it is less clear than what is written currently

padiwik:
i may have misinterpreted the convention... Suppose Alice has two 1s, one of which is known to be red and the other known to be blue (same number of positive color clues). If she plays the wrong 1, this indicates an order chop move instead of a priority?

Zamiell:
in a no variant game?

padiwik:
yes

Zamiell:
```
* Sometimes, after getting clued about two or more 1s, a player will receive a follow-up clue that "fills in" one or more of the 1s. In this situation, playing a 1 with the full identity revealed will not cause an *Order Chop Move*. In other words, *Order Chop Moves* can only happen when a player has two or more 1s with each 1 having two or more identities.
* Besides that, there are some other common situations that do **not** imply an *Order Chop Move*. Use your common sense! Here is a short list:
  * The 1 was expected to be played from the *[Fresh 1's Rule](#playing-multiple-1s---the-fresh-1s-rule-part-2)*.
  * The 1 was expected to be played from the *[Chop Focus Exception](#playing-multiple-1s---the-chop-focus-exception-part-3).
  * The 1 that was played has different positive clues than the other 1's.
    * Different negative clues are ignored.
  * The 1 that was played was originally *Finessed* from a clue that happened **before** the number 1 clue.
  * The 1 that was skipped over is known to be bad from contextual information. (See the *[Stale 1's Clue](#context-clues--the-stale-1s-clue)* section.)
```

padiwik:
ok let me rebase off that

padiwik:
check now

PR #354, opened 2020-09-04
Update Beginner.md - advanced Prompt example.

bnordli:


Beginner.md
```diff
@@ -304,14 +304,17 @@ In this guide, there will be some screenshots to show off the concepts introduce
...
   * Bob has three separate unknown 3's in his hand.
   * Alice clues red to Cathy, which touches the red 4. This is a *Play Clue* on the red 4.
-  * Bob must have the red 3, so this is a *Prompt*. Bob plays his left-most 3 from slot 2.
+  * Bob sees that it is not playable and concludes that this is a *Prompt* and he must have the red 3. Bob plays his left-most 3 from slot 2.
   * However, after playing it, Bob sees that it is not the red 3 - it was the blue 3 (which played successfully). Bob now knows that the next-left-most 3 must be the red 3 (on slot 3).
+  * Since Bob played a card to make her card playable, Cathy now knows she must have the red 4. (She can see the red 3 in Bob's hand and knows he will eventually play it.) She does something unrelated.
   * On his next turn, Bob tries to play his slot 3 card as red 3. However, after playing it, Bob sees that it was not the red 3 - it was the green 3 (which played successfully). Bob now knows that his final 3 must be red 3 (on slot 4).
   * On his next turn, Bob plays his slot 4 card, and it is red 3.
+  * Cathy can now finally play her red 4.
+* This was a very efficient 4-for-1 clue, and a good play by Alice.
```
Zamiell, line 317: this is not a 4 for 1, it is a 1 for 1

bnordli: Because they were not played blind, right? Is this wording better?

Zamiell: if the cards are already clued then it doesn't get them. they are already gotten via gtp

Zamiell: https://github.com/Zamiell/hanabi-conventions/blob/master/Reference.md#efficiency

Zamiell:
>   * Since Bob played a card to make her card playable, Cathy now knows she must have the red 4. (She can see the red 3 in Bob's hand and knows he will eventually play it.) She does something unrelated.

this is not true. Cathy knows she has the red 4 because of GTP, and this is independant/regardless of what Bob does, so this is written in a misleading way.

> * Four cards ended up being played by this clue, which is very efficient, and a good play by Alice.

just delete this line. a 1-for-1 clue is **not** very efficient. in fact, it is actually the most minimum efficiency clue you can do in the entire game without it being illegal (MCVP)

bnordli:
I see, but then this section under Special Moves is wrong/misleading as well, since it doesn't really apply to Prompts, only Finesses.
> There are two special "moves" that players can do in order to get multiple cards played by only using 1 clue, which is very efficient.

Zamiell:
yeah, lets remove that too, and have it just say "There are two special "moves" that players can do in order to get multiple cards played by only using 1 clue, which can be very nice for the team."

Zamiell:
thanks

PR #355, opened 2020-09-05
Specify further the rules of 5 Pull Finesses

timotree3:
This is just my working understanding and I would love feedback.

I also have one question. There's a scenario that some experts say is legal but would conflict with this wording.

- Cathy has one clued 1 in her hand - the red 1.
- Donald has a red 2 on finesse position
- Alice does a *5 Pull* on Bob's red 3
- Alice expects Bob to wait for red 1 to play before blind-playing.
- Bob discards.
- Cathy plays red 1
- Donald blind-plays red 2
- Alice discards
- Bob blind-plays red 3

Is Bob required to delay his blind-play even if he knows that his pulled card cannot be red 2? For example, if he sees the other copy in Alice's hand

If he is required to delay it, then this clue by Alice seems perfectly legal. In which case the wording used in this commit ".., must be demonstrated with a blind-play before the pulled player's next turn" is incorrect.


Zamiell:
seems good

PR #362, opened 2020-09-22
update bad cm example

Aigul9:


Zamiell:
i fixed it without changing it to yellow

Aigul9:
zam suppresses creativity :/

PR #368, opened 2020-10-05
Fix typos

vEnhance:
Repair some broken hyperlinks, fix a "does does not"

(There were some other typos I saw but didn't note down, so I can't remember where they were now... will keep making pull requests as I find them :P )


Zamiell:
thx

PR #373, opened 2020-10-24
Proposal: Removed levels 12-13 and replaced with Appendix A - 5 Tech

sephirothx:
Since levels 12-13 contain some very straightforward moves that are very easy to grasp once players get to level 8+, more and more players are using them before getting to the harder levels like 10-11.
I think it makes sense to group them under an Appendix so that people can say they play "Level 9 + 5 tech" instead of "Level 9 + 5ce + 5pull".

### DIFF
--- a/Learning_Path.md
+++ b/Learning_Path.md
@@ -152,9 +152,9 @@ The following is a brief description of each level.
 
 #### Level 12 - Senior
 
-* Level 14 strategies should only be learned if you have 240+ games of experience with the group.
+* Level 12 strategies should only be learned if you have 210+ games of experience with the group.
 * This level covers moves that are complicated, difficult to see, or require a deeper understanding of how Hanabi works in order to work properly.
-* Level 14 strategies are as follows:
+* Level 12 strategies are as follows:
   * *[The Occupied Finesse](Reference.md#the-occupied-play-clue--the-occupied-finesse-opc)*
   * *[The Trash Push](Reference.md#the-trash-push)*
   * *[The Trash Finesse](Reference.md#the-trash-finesse-strong-form)*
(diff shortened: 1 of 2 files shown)


Zamiell:
i dont want to do an appendix because it makes it more complicated.
just dont do 5 pulls unless you are actually a level 13 player.
this is thoroughly explained in the first section of the learning path document

sephirothx:
Just a proposal to make the learning path document reflect better the real way people are playing. People will keep playing this way whatever the document says anyway.

Zamiell:
maybe you are confused as to what the learning path is for.
the point of the learning path is to convince people to stop using high level moves in lower level games.
maybe the learning path can be tweaked here or there, and maybe some moves can be shuffled around between levels.
but more generally, if people are using high level moves in lower level games, I don't think that is a problem with the learning path.
if that is the "real way that people are playing", then it means that:
1) high level players are being **selfish** and not playing in an appropriate way with lower level players.
2) lower level players are being overconfident and playing with moves that they are not ready for yet.

timotree3:
It seems to me that new players are just practicing the conventions in the order they find most interesting rather than the order presented in the document. Does that have anything to do with overconfidence?

What do you think about documenting what you intend to achieve with the order and division of the levels in a file similar to `Convention_Goals.md`? That seems like it would be informative to new players choosing their learning path as well as to players like Sephiroth who want to propose changes.

EhDrone:
Sounds like there's two discussions going on now - - Zamiel is stating the point of the doc isn't to appease how people are actually playing, but instead to try and push a standard learning path for people to follow so that everyone at higher levels actually play at that level. 'Interesting' isn't an argument to re-arrange things, as a pestilent double bluff is certainly more 'interesting' than a 5-pull, but I don't think you would make the argument that it should be moved out of 17+ just because its a 'cooler' move right? And it certainly has to do with overconfidence - the number of fresh people with like 20 something games who say that they are "level 10" or "level 13" is a bit astounding, and like clockwork they all mess up simple things like a simple finesse or even save clues.

Perhaps timo + seph you guys would rather make the argument that the 5 moves should be moved to a lower level in the doc and they don't belong in 12+ for another logical reason related to the learning path, the complexity of the move, or its relationship to other conventions?

vEnhance:
>  maybe the learning path can be tweaked here or there, and maybe some moves can be shuffled around between levels.

For the record I would also cast my vote for simply moving 5ce's (currently lvl12) and maybe 5pull's (currently lvl13) before priority (currently lvl10) and elimination notes (currently lvl11). I personally struggled a lot with the latter two compared to the former two --- the basic 5 moves seem easier to "see" whereas I still miss priority all the time.

(But I also understand it's confusing for the levels to change on a whim.)

Zamiell:
> It seems to me that new players are just practicing the conventions in the order they find most interesting rather than the order presented in the document. Does that have anything to do with overconfidence?

whether or not a move is interesting is irrelevant for the purposes of this discussion.

if:
- a new player disregards the learning path
- and instead learns whatever moves that they find interesting, then
- they will be much more difficult to play with than someone who is "a level 4 player"

this is for a multitude of reasons that go far beyond the simple fact that you can abbreviate the conventions that the latter is comfortable with in a verbally concise way.

namely:

> the number of fresh people with like 20 something games who say that they are "level 10" or "level 13" is a bit astounding, and like clockwork they all mess up simple things like a simple finesse or even save clues.

a hundred times this.

so maybe "overconfidence" is the wrong word, but it sounds to me like you are just discussing/arguing semantics.
you could replace "overconfidence" with "annoying", "harder to play with", "selfish", "not conducive to playing in a team environment", etc.

> What do you think about documenting what you intend to achieve with the order and division of the levels in a file similar to Convention_Goals.md?

i think that the purpose of the learning path is clearly explained in the introduction of the learning path, using some exposition to really drive the point home.

if you don't think that what I wrote is clear, then submit a pull request

sephirothx:
> For the record I would also cast my vote for simply moving 5ce's (currently lvl12) and maybe 5pull's (currently lvl13) before priority (currently lvl10) and elimination notes (currently lvl11). I personally struggled a lot with the latter two compared to the former two --- the basic 5 moves seem easier to "see" whereas I still miss priority all the time.

They are pretty easy to grasp and see, and mostly people don't have a way to make someone play their 2nd finesse position before level 12 and they ask if there's a way to do that - and at that point just teaching about the existence of 5ce fits much better in the natural learning process rather then just saying "get to lvl12 and you'll find out"

> (But I also understand it's confusing for the levels to change on a whim.)

I completely agree, that's why I thought an appendix would be a cleaner way to achieve this rather than shuffling the levels around

Zamiell:
as previously stated, i'm not opposed to shuffling the levels around, and would rather do that than add an appendix

Zamiell:
do you want to do a PR that moves 5pull to lv 10 and and 5CE to lv 11 and moves everything else upward

vEnhance:
> do you want to do a PR that moves 5pull to lv 10 and and 5CE to lv 11 and moves everything else upward?

OK, let me take a stab at it.

rz-1:
>
>     1. high level players are being **selfish** and not playing in an appropriate way with lower level players, or
> 
>     2. lower level players are being overconfident and playing with moves that they are not ready for yet.

This is accurate based on my experience as a new player playing on the site for 2 weeks. I would break #1 down into two parts a) players giving these clues as a friendly teaching moment and b) players giving strange clues and talking down to you during the review expecting you to respond accurately to them. 

I'm also guilty of #2.

Zamiell:
sounds like you might have run into some bad apples.
some people are good teachers and some are not so good.
i'm sure that noone is intentionally trying to be a dick, and i like to give people the benefit of the doubt where possible.
a lot of conflict can be avoided by clear, verbal communication
e.g. before the game starts:, "i am a level 4 player and am still learning the game, i will try my best but please hold off on using more advanced moves for now".

PR #377, opened 2020-10-26
Proposal: Shuffle levels so 5ce + 5pull come before prio + elim

vEnhance:
Implement the suggestion in last comment of #373


Zamiell:
i think you forgot table of contents, correct me if i am wrong

vEnhance:
> i think you forgot table of contents, correct me if i am wrong

Assuming you're referring to ToC at start of `Reference.md`, I don't think that table of contents needs to change because I left the names of the levels intact (journeyman, advanced, adept, .senior).

Zamiell:
you're right

Zamiell:
thanks

PR #378, opened 2020-10-26
Proposal: Moved 5 Tech to lvl 10 and pushed the other levels down

sephirothx:


sephirothx:
Let me know if there is anything you would like to change @Zamiell

Zamiell:
looks like you and venhance both submitted a PR doing the same thing, but he was first, so ill accept his PR. thanks

sephirothx:
I had also merged 5ce and 5 pull in a single level because it seemed overkill to have a specific level for 5ce alone, but it's fine as long as they come before priority

PR #383, opened 2020-10-27
Update Reversed.md

pianoblook:


Zamiell:
im already fixing this

PR #390, opened 2020-10-28
Prompt bluff example fixed

DrAnax:
Since Bob played his clued 2, Cathy shouldn't play because Bob doesn't see red 2 on her.

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -1449,9 +1449,9 @@ Note that in this user interface, players are **not** holding their hands like t
   - Red 1 and blue 1 are played on the stacks.
   - Bob has a clued 2 in his hand and does not know its true identity.
   - Alice clues red to Donald, touching a red 3 as a *Play Clue*.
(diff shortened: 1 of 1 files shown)


Zamiell:
> Since Bob played his clued 2, Cathy shouldn't play because Bob doesn't see red 2 on her.

wrong

> Cathy is slightly surprised that Bob played the blue 2, but he must have been able to figure it out somehow

Zamiell:
meaning that sometimes players can play their clued cards, this isn't super unusual

DrAnax:
Wait,
> Bob knows that this must be a Prompt, so his 2 must be red 2. He plays it, and it is instead the blue 2.

This can't be true if Cathy has red 2.

DrAnax:
Ok melwen explained better this to me. Cathy would play if Bob doesn't have red 2 on his finess position and Bob would play if he doesn't see red 2 on Cathy.

PR #392, opened 2020-10-28
examples pr

iamwhoiamhahaha:
Examples for Bubblegum Wraparound, 5 Pull Double Finesse, and Bad Chop Move Ejection.
(Should I upload the screenshots when the clue was given or before the clue was given?)


Zamiell:
before the clue was given

Zamiell:
you also have to add the links in the main reference.md and add the example.md documents with all the text explaining, matching the format used in the other example.md documents

PR #409, opened 2020-11-11
Fix typos in lvl17 master addition

vEnhance:
* Broken link in reference
* Add header row to table so it renders correctly in learning path


vEnhance:
BTW, while reading through Pestilent Double Bluff examples, I noticed there were still a bunch of references to Disease Bluffs, which are a recently deleted convention. I think this is a mistake, so I cleared those; if I understood incorrectly, feel free to drop the last commit 5e40cd2.

Zamiell:
looks good, thanks for your work evan

PR #416, opened 2020-11-22
Limit Trash Bluff to when there is basically only one card missing

timotree3:
Note that the proposed definition is intended to cause a *Trash Bluff* if negative clues and cards visible in hands eliminate all useful cards that are not playable and match the clue.
e.g.
- The purple 3 is played on the stacks
- Alice clues purple to Cathy touching a purple 3 with negative 5 on it as a play clue
- Bob plays *Finesse Position* as a *Trash Bluff*

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -2312,6 +2312,8 @@ There are a lot of different moves that involve cluing trash. [Here is a handy f
   - Bob blind-plays his *Third Finesse Position*, and it successfully plays as green 2.
   - Cathy knows that a *Discharge* has occurred from Alice's red clue. Thus, this must be an *Unknown Trash Discharge*, so Cathy's red card must be trash. Cathy marks her red card as red 1 (since that is the only possible trash red card) and discards it.
 - When multiple cards are clued with an *Unknown Trash Discharge*, **only the focus of the clue is considered to be trash**. This means that *Good Touch Principle* applies to the non-focused cards. (Note that this is **the opposite of the *Trash Bluff***. In a *Trash Bluff*, when multiple cards are clued, **all** of the touched cards are considered to be trash.)
+- More specifically, it is an *Unknown Trash Discharge* if the trash card focused by the clue could be a useful card that is not playable, otherwise, it is a *Trash Bluff*.
+  - This does **not** take into account harder to see factors that would make a *Bluff* impossible, such as that the card cannot be *Bluff Target* or that if the card was a *Bluff Target*, *Bob* would wait for a clued card before blind-playing.
 
 ### 5 Number Discharge (5ND)
 
--- a/Reference.md
+++ b/Reference.md
@@ -2312,7 +2312,7 @@ There are a lot of different moves that involve cluing trash. [Here is a handy f
   - Bob blind-plays his *Third Finesse Position*, and it successfully plays as green 2.
   - Cathy knows that a *Discharge* has occurred from Alice's red clue. Thus, this must be an *Unknown Trash Discharge*, so Cathy's red card must be trash. Cathy marks her red card as red 1 (since that is the only possible trash red card) and discards it.
 - When multiple cards are clued with an *Unknown Trash Discharge*, **only the focus of the clue is considered to be trash**. This means that *Good Touch Principle* applies to the non-focused cards. (Note that this is **the opposite of the *Trash Bluff***. In a *Trash Bluff*, when multiple cards are clued, **all** of the touched cards are considered to be trash.)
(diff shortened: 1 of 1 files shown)


Zamiell:
i dont think this clarification helps the section very much.
in a trash bluff, you clue a card that "could be a useful card that is not playable".
so why do you have that as the definition for a utd

timotree3:
> in a trash bluff, you clue a card that "could be a useful card that is not playable".
No?
in a trash bluff you clue a card that is either playable or trash. it cannot look unplayable and useful and still be a trash bluff. give me an example where what you're saying is true

Zamiell:
y3 is played on the stacks.
alice clues cathy yellow, touching a yellow 3.
from cathy's perspective, this is a yellow 4.
yellow 4 is a useful card. but since it is really a yellow 3, it is not yet playable

Zamiell:
if donald has a y5 in this hypothetical, it is a trash bluff.
if y5 is not visible in this hypothetical, it is a utd

timotree3:
> y3 is played on the stacks.
> alice clues cathy yellow, touching a yellow 3.
> from cathy's perspective, this is a yellow 4.
> yellow 4 is a useful card. but since it is really a yellow 3, it is not yet playable

I mean just looking at the empathy on the card if any possibilities are present that are both useful as well as not playable. In your example the answer is yes (yellow 5) and therefore Bob discharges.

Anyway, take a look at the new wording

Zamiell:
> I mean just looking at the empathy on the card, if any possibilities are present that are both useful as well as not playable. 

that is not a sufficient criterion either
e.g.
red 1 and yellow 1 are played on the stacks.
alice clues 1 to cathy.
bob blind-plays green 1.
cathy's empathy shows that the card can be blue 1 or purple 1

timotree3:
what are you talking about? in that example the possibilities that are useful are green 1, blue 1, purple 1. notice that they are all playable, so the possibilities that are useful and not playable are zero.

Zamiell:
oh i see

timotree3:
Did you look at the new wording though? It's much clearer
> More specifically, it is a *Trash Bluff* if the clued card is known to be either playable or trash, otherwise, it is an *Unknown Trash Discharge*.

Zamiell:
i dont think that is very clear. trash is not playable. and you never initiate a trash bluff with a playable card

timotree3:
Of course Bob knows what the card is. The point is what possibilities Bob sees on the card's empathy

timotree3:
I don't know what words you want me to use besides "if the clued card is known to be" that would signal what the clue receiver sees when they look at the card.

Zamiell:
how about this:

```
- In summary, for a *Trash Bluff*:
  - Before the blind-play, the clued card looks playable. (Just like an *Unknown Trash Discharge*.)
  - After the blind-play, the only card possibilities left are playable cards.
  - Thus, the blind-play is enough to prove it is trash.
- In summary, for an *Unknown Trash Discharge*:
  - Before the blind-play, the clued card looks playable. (Just like an *Trash Bluff*.)
  - After the blind-play, the card possibilities include one or more cards that are currently unplayable.
  - Thus, something additional is needed to prove it is trash.
- Other than this, don't use more complicated factors to prefer a *Trash Bluff* over an *Unknown Trash Discharge* (like Bob having to wait on some other card).
```

timotree3:
No because now it's still about what the card looks like after a *Finesse Position* blind-play, taking into account all of our conventions which causes confusing situations. It needs to be about what the empathy on the card shows before the blind-play and nothing else

Zamiell:
```
- In summary, for a *Trash Bluff*:
  - The empathy on the clued card **only** contains possibilities that are all *Playable* or *Delayed Playable*.
  - Thus, a normal blind-play is enough to prove it is trash.
- In summary, for an *Unknown Trash Discharge*:
  - The empathy on the clued card contains one or more cards that are currently unplayable.
  - Thus, something extra is needed to prove it is trash.
- Other than this, don't use more complicated factors to prefer a *Trash Bluff* over an *Unknown Trash Discharge* (like Bob having to wait on some other card).
```

timotree3:
Great except that it's slightly wrong because the empathy will also include trash cards.
```
- In summary, for a *Trash Bluff*:
  - The empathy on the clued card **only** contains possibilities that are *Playable*, *Delayed Playable* or trash.
  - Thus, a normal blind-play is enough to prove it is trash.
- In summary, for an *Unknown Trash Discharge*:
  - The empathy on the clued card contains one or more useful cards that are currently unplayable.
  - Thus, something extra is needed to prove it is trash.
- Other than this, don't use more complicated factors to prefer a *Trash Bluff* over an *Unknown Trash Discharge* (like Bob having to wait on some other card).
```

Zamiell:
ok i added that

PR #426, opened 2020-11-30
4 Charm / 1's Chop Move / Self Color Ejection / Deceptive-Fives

iamwhoiamhahaha:
I also wrote the document for SCE since it was my idea.


Zamiell:
thanks

Issue #311, opened 2020-07-13
Ejection/Discharge Theory and Proposals


pianoblook:
Over the weekend I wrote up a [guide](https://docs.google.com/document/d/1n3IjeFZo-VFEf5KSsk9mu_hDjPFvooZjXEuiYk0FSu4/edit?usp=sharing) that aims to unify + simplify all Ejections and Discharges.

Goals and included proposals:

1. A clean definition of Ejections/Discharges that aims to drastically simplify the what and why of these moves

2. A proposal to split the rarely-used _Double Play Ejection_ into two cleaner conventions (one Ejection and one Discharge)

3. 6 "new" opportunities for Ejections and Discharges, that don't interfere with any other conventional interpretations

4. A proposal for _Unnecessary_ forms of the 1-for-1 Ejections/Discharges. These are really just extensions of the [UTCP Proposal](https://github.com/Zamiell/hanabi-conventions/issues/237)

Just start [here](https://docs.google.com/document/d/1n3IjeFZo-VFEf5KSsk9mu_hDjPFvooZjXEuiYk0FSu4/edit#heading=h.b3ar6es9llx4) in the doc if you want to jump to the juicy stuff.

pianoblook:
<img width="1201" alt="Screen Shot 2020-07-14 at 12 20 55 AM" src="https://user-images.githubusercontent.com/32303453/87383750-1bafd400-c568-11ea-8539-0bf084e6f0ef.png">

In-game example of what could have been an Unnecessary version of a _Trash Push Discharge_.

Context: multi-5 is known, b4 is known non-green, and Bonja's m1 is basically known trash (non-4/5)

- With negative green on piano's 4, a simple 4 clue would have gotten it played.
- Instead, cluing Black to Bonja's trash 1 would trigger a _Trash Push Discharge_
- After b4 plays, Bonja wonders why Nitrate would go through all this trouble to get the b4....
- He must be promising something extra: a _Trash Push_ on the card to the left of the 'pushed' card (which is now also known trash)

pianoblook:
Here's a list of the individual Ejections proposed. Some have been proposed in the past, and I'm hoping this unifying 'Signal Shift' theory can help convince folks that these are all super intuitive and fall perfectly in line with existing Ejections/Discharges.

**1. Trash Push Ejection**
-	If a _Trash Push_ is given on a card that would require the next player to have two or more blindplays to fulfill, this calls for an Ejection.
[...]

Issue #257, opened 2020-06-02
Trash Push Ejection Example


iamwhoiamhahaha:
1 to charmander as a *Trash Push Ejection*

![trash push ejection example](https://user-images.githubusercontent.com/50160641/83522339-1dbb5780-a513-11ea-90e7-78fc91d2ff06.png)

pianoblook:
What are your thoughts on restricting the _pushed_ card to be any >1-away-from-playable Critical card (and therefore letting it be 'cm'd for free)?

Here I see two other ejections possible anyway: 5ce for sure, but also even a double play clue ejection. Or just cluing the 4 directly, lol

NoMercyO:
Black to charmander as 5CE also valid, and much better?

Zamiell:
5ce seems like the optimal move.
find a better example where TPE is actually the optimal move.

iamwhoiamhahaha:
what if charmander doesn't have a k5 lol. okay but i agree this is not a good example

PR #448, opened 2020-12-23
Proposal: Trash Push Charm, OOP Charm

0nen:


pianoblook:
If the _Pushed_ card is two or more away, why wouldn't this just call for an Ejection? 

Your example, Alice cluing red to Cathy, is not a trash push, because it is not already known Trash. Bob would need to Discharge and trust that it was a _Necessary UTD_

See #257 and #311

pclinger:
In the example, red 5 is played on the stacks, so clueing red is known trash, and thus a trash push.

0nen:
In the example, red 5 is played on the stacks. A clue of red is most definitely a trash push.

pianoblook:
Oh ok. well in that case it would not conflict with 5ce because it's already known trash

Jayhui-q:
Agree with Piano's comment above. If it's known trash, the ejection is not a lie and is a clearer signal shift of play -> save/good card, and it is probably more likely that an ejection exists and is playable than for a charm. For 4 players or more you can always focus the charm slot if it exists, and you wouldn't be worried about the aforementioned good card being chopped.

I would be interested maybe if there's some always unnecessary component that gets two blind-plays from some clever maneuver like this, but otherwise, it doesn't seem that strong.

0nen:
There are so many ways to eject cards already, and right now there's no good way to Charm in the mid game. Once all the 1s are played, 4 Charm is gone, and Declined 5 Pull is only in the Low Score Phase.

Having more tools opens up options up for new play. Many times you are stuck in a position where you must 1-for-1 clue a chop card that is about to fall off. Something like this adds a scenario where you can still get that 4th finesse to play, and save another useful card, for later.

I feel like a trash push on a two-away-from-playable card is a scenario that could come up quite often in games. This could make it a very valuable tool for getting 4th finesse position, where we just don't have that option right now.

Jayhui-q:
Haha, I think you would love to see the Junk Charm that Piano proposed then: 
https://github.com/Zamiell/hanabi-conventions/issues/399

I agree about having the new tools, but the charm moves the signal from play to play rather than from play to save. 
https://github.com/Zamiell/hanabi-conventions/blob/master/Reference.md#level-19---rarely-used-strategies---signal-shifting

In addition, in your example, an ejection would never be a lie. If it is known trash, ejecting does not make it read as a 5 (especially if you give a rank clue, which your proposal supports), and if it were unknown trash, then a finesse/discharge would be needed.

The reason we have more ejection tools than charm tools is precisely that the charm slot is less likely to exist or be playable. (A player with 4 cards in hand must have no cards clued at all for it to exist, and if we've let it go all the way to the charm slot, it must've not been playable for a long time). This is also why it makes sense that most of our charms happen in the low score phase.

I agree that a trash push on a two-away-from-playable comes up a lot, but I think the argument is that it should eject instead. This would protect the card on the chop that is about to fall off too! Otherwise, the usual solution to a card on slot 4 is just to give the save and then prompt it out later, which I don't think is that bad.

pianoblook:
Trash Push Ejections have been proposed for many months.

This discussion from back in April (plus Jeff's linked doc) is a treasure trove of discussion

https://github.com/Zamiell/hanabi-conventions/issues/186

Zamiell:
open an issue for convention proposals, not a pull request.
i will ask for a pull request if the proposal is accepted

PR #454, opened 2020-12-31
Update Beginner.md

Pimak:
Numbering was off by one in the document.
It still looked good, but it bugs me to know that there are two 1s


Zamiell:
they should actually all be 1s like it is in Reference.md

PR #458, opened 2021-01-03
Modify Efficiency explanation

periodically-makes-puns:
The site now shows future required efficiency, not current efficiency. As such, the level 3 explanation of [Efficiency](https://github.com/Zamiell/hanabi-conventions/blob/master/Reference.md#efficiency) needs to be modified to reflect this change.


Zamiell:
thanks

PR #461, opened 2021-01-04
Clarify Delayed Play Clues

padiwik:
There was a beginner confused about this. The example was not very concrete beforehand, anyways.


Zamiell:
don't use yellow

the convention in the doc is to use red, then blue, then green

padiwik:
you mean red, green, blue?

Zamiell:
nope, red, blue, then green

padiwik:
Huh, why? But fixed

Zamiell:
thats the order that colors are assigned on the website

padiwik:
Oh yeah because a game with 2 normal suits is just red and blue, right

Zamiell:
yup

Zamiell:
i updated the doc to be more clear

PR #463, opened 2021-01-05
Clarify scream discard description

scmccarthy:
There was a possible misinterpretation of scream discard, namely that it refers to discarding the playable card specifically, rather than generally taking a "discard" turn. Several sentences about it were written ambiguously.

Edit those sentences to remove the ambiguity.

### DIFF
--- a/Reference.md
+++ b/Reference.md
@@ -850,7 +850,7 @@ Note that in this user interface, players are **not** holding their hands like t
 
 - See the *Chop Move* section above for an explanation on what a *Chop Move* is.
 - The *Chop Move* outlined above is a tool to save multiple cards with one clue. However, sometimes you may not have a clue at all. One way to solve this situation is with a special kind of discard.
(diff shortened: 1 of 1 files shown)


hanabirnoob:
I remember when I first made this mistake.

padiwik:
Interesting . Did your brains read "discarding instead" as "discarding it instead"?

willwin4sure:
Might as well say "discard your chop instead" to be more clear?

scmccarthy:
> Interesting . Did your brains read "discarding instead" as "discarding it instead"?

No - at the point of that sentence it is still likely that it's talking about discarding a different card. However, then it continues:

> This is called a *Scream Discard*. Since your teammate expected you to play the card, it is like screaming at them to let them know that things are very bad.

Writing "the card" there suggests that there is only one card involved, rather than two different ones, the first of which is the referent.

Zamiell:
looks good, thanks

PR #477, opened 2021-01-25
Loaded white play clues

rkass:
Loaded White Play Clues. Expands the ability to give loaded play clues to white cards on chop at the cost of limiting the ability to give loaded save clues.

This would've been useful here: https://hanab.live/shared-replay/384380#14


Zamiell:
if you are proposing a new convention then you need to do that in an issue, not a PR

PR #488, opened 2021-02-07
Cursed, Prophetic, and Signal Shift pruning

pianoblook:
I didn't know where to put the updated Signal Shift graphic, since the image links seem to be busted atm


pianoblook:
signal_shifting

Posting it here just in case

Zamiell:
the image links seem to be working fine: https://hanabi.github.io/docs/level_19/signal_shifting/
not sure what you are talking about

PR #521, opened 2021-02-28
Generate example images programmatically.

DarthGandalf:
I'll leave a better integration of this into the build system to you.


Zamiell:
wtf, rebase your branch on master

Zamiell:
oh, nm it is

Zamiell:
ok lets try it out

Zamiell:
im working on generating the svg cards, it isn't easy

Zamiell:
the appearance of cards has been updated since the SVG commit that i found, so i have to backwards integrate it

Zamiell:
why did you create elements for each image instead of just doing `Poop` ?

DarthGandalf:
> why did you create elements for each image instead of just doing Poop ? e.g. the markdown way to show images

That's what I tried at the beginning. But then `<image>` with the "sprites" of cards weren't showing, because `` is supposed to be self-contained. The external images were silently not loaded from the website. When embedding `<svg>` directly in the resulting HTML, it wasn't a problem anymore.

PR #529, opened 2021-03-02
Update level-3.md

heff10:
Swap Playing 1's images


Zamiell:
the fix is to rename the files, not swap the order so that they are out of order

PR #545, opened 2021-03-05
Regen images when py changes, assign colors to focus and chop

DarthGandalf:


Zamiell:
very nice, thank you so much

PR #557, opened 2021-03-07
Show big text for bluff/finesse

DarthGandalf:


docs/image-format.md
```diff
@@ -23,6 +23,9 @@ import CardTextResult from '@site/image-generator/yml/generator-docs/card-text.y
 import TrashSource from '!raw-loader!@site/image-generator/yml/generator-docs/trash.yml';
 import TrashResult from '@site/image-generator/yml/generator-docs/trash.yml';
```
Zamiell, line 25: no space between imports

DarthGandalf: I disagree. Look above.
These imports are paired source/result

Zamiell:
thanks

docs/image-format.md
Zamiell: ok, i take it back

PR #574, opened 2021-03-10
Update level-6.md

wjomlex:
Both Discard Modulation examples refer to an Early Game exception from Level 8. I'm removing one example, and changing the other to a normal Mid Game situation. I'll update the image next.


Zamiell:
thx

PR #580, opened 2021-03-10
Create svg fix

DrAnax:


DrAnax:
I want to use empty text to leave some space, because the next player's clued cards get too close to the text and overlap it

DrAnax:
image

vs

image

Zamiell:
ok, makes sense!

PR #590, opened 2021-03-11
Decrease size of shadow to make color pips (blue, purple) look better

DarthGandalf:
Dump all the filters and gradients unconditionally - SVGR removes all
unused ones anyway


Zamiell:
thx

Issue #502, opened 2021-02-25
Turn unnecessary trash bluff / UTD into true trash push


florian-5f3759df:
**Currently:** When an unnecessary trash bluff / UTD is performed using a card on Donald's chop, it asks for a single extra blind play. If Donald has something playable on his new chop, it's a trash push, otherwise Cathy plays slot 1.

**Proposal:** When an unnecessary trash bluff / UTD is performed using a card on Donald's chop, it becomes a trash push.

**Implication:**
- If the trash pushed card is playable, Cathy does nothing.
- If the trash pushed card is one-away, Cathy plays slot 1 (like a normal _Trash Push Finesse_).
- If the trash pushed card is 2+ away, Cathy ejects (_Trash Push Ejection_).
- If the trash pushed card is trash, Cathy discharges (_Trash Push Discharge_).

If we need a name, we could call it an Induced Trash Push, and then an Induced Trash Push Finesse etc. (Or some other adjective would work too. Extra? Added? Generated?) 
 
In the original proposal by Kakashi and me, we already wanted to configure it like this. At the time, the unnecessary stuff was already a big change, so we kept backwards compatibility with trash double bluff to make it more palatable. Now trash double bluff doesn't exist anymore, and we have double ignitions to get double slot 1 if we want. 

**Example** by Kakashi:
Played on the Stack: g1 & b3
Alice clues green to Donald, the r1 in Bob's hand could've easily been clued and is not blocked
Bob's hand: X - X - r1 - X
Cathy's hand: X - X - r2 - X
Donald's hand: X - X - b2 - g1
Old configuration: Cathy would play slot 1, and Donald marks his b2 as maybe-trash, maybe-good.
Proposed change: Both Bob and Cathy play slot 3, Donald marks b2 as known trash.

In my opinion, the proposed configuration is actually simpler and easier to explain than what it currently is. Namely it can be summarized as 'treat it like a trash push'.

I seem to remember maybe some teams already play like this, so in that case: please share your experience.

pianoblook:
This fits exactly with my proposal for _Unnecessary Trash Compositions (with Off-Chop Trash)_ in the google doc from #479 

I would be a huge fan of both of these changes

Dr-Kakashi:
Example by Lel0uch:

Played on the stack: G4, R1 and O5
Bob's Hand
r2 - r1 - r5 - r1
Cathy's Hand
r3 - b1 - b3 - b4
Donald's Hand
[...]

PR #608, opened 2021-03-14
Turn unnecessary trash bluff / UTD into true trash push, update level 23

florian-5f3759df:
Turn unnecessary trash bluff / UTD into true trash push, issue #502

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -15,6 +15,7 @@
         "asddsa",
         "ated",
         "bauza",
+	"bcme",
         "bga's",
         "browserslist",
         "carunty",
--- a/.cspell.json
+++ b/.cspell.json
@@ -15,7 +15,7 @@
         "asddsa",
         "ated",
         "bauza",
-	"bcme",
+        "bcme",
         "bga's",
         "browserslist",
         "carunty",
--- a/docs/level-23.md
+++ b/docs/level-23.md
@@ -33,7 +33,7 @@ import TrashPush from '@site/image-generator/yml/level-23/trash-push.yml';
 ### Unnecessary Moves with Known Trash --> Ignition
 
 - If **known trash** is used as clue target, the "extra" thing is an *Ignition*. The *Ignition* can be on anyone on the team, similar to how *Double Ignition* works.
-  	- Applicable conventions include *Trash Push*, *Trash Push Finesse*, *Trash Pull*, and performing a *Trash Chop Move* on a card which can be clued directly.
+  - Applicable conventions include *Trash Push*, *Trash Push Finesse*, *Trash Pull*, and performing a *Trash Chop Move* on a card which can be clued directly.
 - For example, this would be an *Unnecessary Trash Push* in a 3-player game:
   - All of the 3's are played on the stacks.
   - Bob's hand is completely unclued.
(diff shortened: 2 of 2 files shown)


Zamiell:
fails linter

florian-5f3759df:
Pretty harsh spell check, it didn't like BCME as abbreviation. Want me to remove it, or is there a way to add it to dictionary?

Zamiell:
right click --> add to workplace dictionaries

florian-5f3759df:
Already did it manually in the dictionary file, hopefully that's the same

.cspell.json
```diff
@@ -15,6 +15,7 @@
         "asddsa",
         "ated",
         "bauza",
+	"bcme",
```
Zamiell, line 18: bad whitespace

florian-5f3759df:
sorry for the hassle, that's a very silly mistake, I'll redo

docs/level-23.md
```diff
@@ -33,6 +33,7 @@ import TrashPush from '@site/image-generator/yml/level-23/trash-push.yml';
 ### Unnecessary Moves with Known Trash --> Ignition
 
 - If **known trash** is used as clue target, the "extra" thing is an *Ignition*. The *Ignition* can be on anyone on the team, similar to how *Double Ignition* works.
+  	- Applicable conventions include *Trash Push*, *Trash Push Finesse*, *Trash Pull*, and performing a *Trash Chop Move* on a card which can be clued directly.
```
Zamiell: i'd rather not include explicit enumeration of moves here since
1) it strikes me as being unnecessary (no pun intended)
2) it has the potential to get out of date when we add new moves

docs/level-23.md
```diff
@@ -48,9 +49,16 @@ import TrashPush from '@site/image-generator/yml/level-23/trash-push.yml';
 
 <br />
 
-### Unnecessary Moves with Unknown Trash (Off-Chop) --> Chop Move
+### Unnecessary Moves with Unknown Trash
+- If Bob already has to play to tell the clue receiver that the clue target is trash, as an "extra thing" the team also interprets the clue as the **known trash** version of the clue.
+	- Applicable conventions here are the *Trash Finesse*, *Trash Bluff*, *Unknown Trash Discharge (UTD)*, *Bad Chop Move Ejection (BCME)*, and *Trash Push Discharge*.
```
Zamiell: same here

florian-5f3759df:
I did it in editor that automatically does tabs to spaces / indent, and then while merging I changed things on the fly in browser too, forgetting that feature is not in text box... Should be okay now I think.

docs/level-23.md
florian-5f3759df: I thought it added clarity, because maybe it's not obvious when first learning whether BCME and trash push discharge are in the known trash set (since the 'focus' of the actual clue is trash) or unknown trash (since Bob needs to blind play to show the team the card next to the known trash is also trash).

Zamiell: i think if you are level 23 it should be obvious, otherwise you are not actually level 23 and you need to play more games at lower levels

Zamiell: one example of each thing is shown in the subsequent example in each section, that is probably good enough

docs/level-23.md
```diff
@@ -102,7 +101,6 @@ import TrashPush from '@site/image-generator/yml/level-23/trash-push.yml';
   - Cathy sees that Donald is about to play green 1 as a *Trash Push*, but green 1 is a trash card, since it's already played.
   - Thus, Cathy plays her third finesse position as a *Trash Push Discharge*.
 
-
```
Zamiell: each section needs a <br />

docs/level-23.md
florian-5f3759df: I meant when BCME is used unnecessarily, there's no example of that at all.
All 2s played.
Cathy has X - Y - r2 - g1
Alice clues 2 to Cathy, Bob ejects a card that could have been gotten directly. The categorization of 'known trash' or 'unknown trash' determines whether Cathy plays X or Y.

florian-5f3759df: Anyway, if you still think it's superfluous I'll delete the lines.

Zamiell: then add an example

florian-5f3759df: okay, I'll rewrite one of the examples

florian-5f3759df:
Removed the applicable conventions line, and added an example which clarifies that bad chop move ejection is unknown trash, instead of known trash.

Zamiell:
thanks

PR #618, opened 2021-03-21
Reorder bluffs to have 3 more together

DarthGandalf:


Zamiell:
naw critical bluff should go after 3 bluff

PR #642, opened 2021-04-17
Add support of simple variants suits to image generator

DarthGandalf:


.cspell.json
```diff
@@ -111,6 +111,7 @@
         "rnrgames",
         "romain",
         "rubbo",
+	"rygbp",
```
Zamiell, line 114: wrongly indented

DarthGandalf: Fixed

PR #654, opened 2021-05-05
Small grammar correction

NumerateClutter:
Level 16, Elimination

"you might think that the team ma**k**e a mistake" -> "you might think that the team ma**d**e a mistake"


Zamiell:
thanks

PR #655, opened 2021-05-05
Fix incorrect player name

mly32:
### In Level 10. 

In one of the 3 player game examples, a player is called `Donald`, when it should really be `Cathy`.


Zamiell:
if it is cathy then why would cathy play the r2? that doesn't make any sense

mly32:
Here is the original snippet:

- For example, in a 3-player game:
  - No cards are played on the stacks.
  - Alice clues red to Donald, touching a red 2 as a *Play Clue*.
  - Bob does not see any red 1's in anyone else's hand. Thus, Bob knows that Alice intends for him to blind-play a card.
  - From Bob's perspective, it can be either a *Finesse* (on exactly red 1) or a *Bluff* (on the 1 of any of the other suits).
  - Cathy has a critical 5 on chop.


For a 3 player game, can there be a `Alice`, `Bob`, `Cathy`, and `Donald`?

From the context, Alice is in the bluff seat of Bob and since it's a 3 player game, the only other person is Cathy. Here, `Donald` should be `Cathy`... should it not?

timotree3:
Donald is intended. It should just say 4 player game.

mly32:
Ahh. That makes so much sense now.

timotree3:
No. In a 3-player game, Cathy will not discard her 5 if the red clue was a *Finesse*, which may mislead player's into thinking that *Guide Principle* means that they should always assume a *Finesse* if there are important cards on chop. In fact, what it means is that they should assume it's safe to blind-play. This is completely clear if Donald is receiving the clue.

mly32:
Your example makes sense. Thanks.

PR #656, opened 2021-05-05
Restrict White Loaded Play Clues to be more reasonable

timotree3:
This is a follow-up to [some recent discussion in Discord](https://discord.com/channels/140016142600241152/456214043351580674/839300317496737843). The current wording may mislead player's into thinking they should assume a chop focus loaded play clue with rank is always on a white card.

I saw in the original discussion that the restriction on "the white card of that rank is playable" was removed intentionally, but I think that makes the White Loaded Play Clue convention too extreme. This is a proposal that is more in line with how I understood the spirit of the convention.


vEnhance:
>    - Note that it could still be a *Play Clue* on a card that isn't white.

Maybe it would be good to update the immediately following example as well to emphasize this. Say, blue 3 in trash, and both yellow 2 and white 2 are played. If so, when Bob receives a number-3 clue on chop, he marks it as either yellow 3 or white 3, but not blue 3.

vEnhance:
Two questions about "No Early Saves on White Cards".

* Are early saves on white 5 permissible in early game? Well actually, I guess that's the only case that makes sense unless a white card is misplayed, but I think this is a change from before.
* Do we permit bluffing (or charms, etc.)? If so, maybe better to write "promise" or similar rather than finesse specifically.

timotree3:
> Maybe it would be good to update the immediately following example as well to emphasize this. Say, blue 3 in trash, and both yellow 2 and white 2 are played. If so, when Bob receives a number-3 clue on chop, he marks it as either yellow 3 or white 3, but not blue 3.

Done.

> Are early saves on white 5 permissible in early game?
 
I would be open to allowing them. As written they are not.

> Well actually, I guess that's the only case that makes sense unless a white card is misplayed, but I think this is a change from before.

No. The main use case for this convention is giving play clues to white 2's on chop.

> Do we permit bluffing (or charms, etc.)? If so, maybe better to write "promise" or similar rather than finesse specifically.

Done

vEnhance:
Ah right, I forgot about white 2.

I think my inclination is to allow early saving white 5 but I don't feel strongly.

timotree3:
I guess it's kind of illogical to allow 5 stalls with white 5 but not allow early 5 saves with white 5. By that thinking, the conventions should say something like "early 5 saves take precedence over these conventions when someone is in a stalling situation and could be giving a 5 stall." That sounds too complicated though.

Zamiell:
the branch has a conflict, can you fix it?

timotree3:
done

Zamiell:
i made some of my own edits, please review

timotree3:
I don't like it. The first section opens with "players are not allowed to early save white cards" which is not actually what that section is about. That section is about "players are not allowed to early save with rank when the white card of that is playable". The second section has been changed to include the word "finesse" which is not actually true, given that it can also be a bluff, double bluff, charm. The second section has been changed to say that it can be derived from the first section, which is also not true. Was there something wrong with my original version?

Zamiell:
> I don't like it.

see my new edits

> The second section has been changed to say that it can be derived from the first section, which is also not true.

can you clarify why this isn't true?

timotree3:
Now that the first section includes both directly playable white card assumption and white loaded play clues it's fine. I put them in two different sections because I consider them two different conventions, but that's just style.

PR #659, opened 2021-05-07
Mockup of example puzzle format

vEnhance:
Mentioned in the discussion starting around
https://discord.com/channels/140016142600241152/776176933120507964/840244878511833139

This is a proposal for a format by which we could design a few puzzles/exercises/etc. for each level so people can get understanding of the new conventions they are learning. I imagine this would be the most impactful at low levels, when you give a new player a position and ask them to list all legal clues then choose the best one, or give a clue and ask the reader to interpret it. But, in principle one could design puzzles for any level like this, albeit it's awkward to deal with context.

image

image


Zamiell:
i like this

Zamiell:
1) the branch has conflicts, can you fix them?
2) it looks like you have other stuff in this pr other than just new puzzle docs? shouldn't that go in a separate PR?

vEnhance:
1. done
2. i think those were a mistake and should be gone now

PR #681, opened 2021-07-04
Improve examples for level 18

DarthGandalf:


Zamiell:
why change run.sh?

DarthGandalf:
Because python deps should already be installed after running `install.sh`

Zamiell:
i can't remember tbh

PR #700, opened 2021-07-12
Change wording for delayed play clues definition

vEnhance:
Open for discussion about this one. However, I think that at minimum, there is some linguistic confusion as written about whether "delayed play clues" are a *subset* of play clues or are a *different  kind of clue*.

I think we should prefer the former. Because in Hyphen-ated, all our play clues could be delayed, I think it is much cleaner to phrase it that way. I'm not even sure we should have another term for it.


Zamiell:
sure

PR #703, opened 2021-07-12
Add a puzzle to the beginner's guide that emphasizes color critical 2 saves

timotree3:
This was already covered in puzzle 3, I just thought it could use more emphasis since this is a very common misunderstanding in my experience.


Zamiell:
ok, thanks

PR #707, opened 2021-07-12
Beginner's Guide: Early Game Question 2

jessgwen:
As currently written, Alice can give a legal play clue to Bob's g2 with either 2s or green, but the solution only mentions green. This PR changes Bob's chop to a trash y2 so that green is the only legal clue.


Zamiell:
thx

PR #715, opened 2021-07-16
Make "Determining Focus" section easier to read

0nen:
Also, the "Clue Interpretation" page mentions a "3-step" process, but nothing on this page said anything about a 3 step process. So, this fixes that issue as well.

### DIFF
--- a/docs/beginner/single-card-focus.md
+++ b/docs/beginner/single-card-focus.md
@@ -20,11 +20,11 @@ import ChopFocus from '@site/image-generator/yml/beginner/chop-focus.yml';
 
 ### Determining the Focus: 3 Steps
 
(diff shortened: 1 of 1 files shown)


Zamiell:
i edited it

0nen:
One issue with the text:

> Otherwise, the focus is on the left-most new card.

There is the scenario where the clue touches multiple cards, but none of them are new. I suggest removing the word "new" from Step 3.

Zamiell:
are you looking at the version on master?

0nen:
Sorry didn't see that there was a commit elsewhere on this same topic, I had just reviewed the commits in this request. LGTM

Issue #713, opened 2021-07-16
Beginners Guide: Play Clues (Question 1 & 2)


0nen:
https://hanabi.github.io/docs/beginner/play-clues-question-1

Q1:
> Bob marks down the identity of the card.
> What identity does Bob write down?

Q2:
> Bob marks down the identity of the card.
> What identity does Bob write down? (Be specific.)

Up to this point there has been no discussion of how you "mark" cards. I suggest rewording the question to simply ask what Bob should think the card is based on the clue.

Edit:

This affects other questions as well, I just haven't read through the whole beginners guide, for example https://hanabi.github.io/docs/beginner/delayed-play-clues-question-1 has the same issue. There are probably more.

Another option is to introduce the concept of marking cards early on.

https://hanabi.github.io/docs/beginner/critical-save-question-1 
On this question, it asks
> What identity does Bob write on the card?

Similar issue here, inconsistent wording (variations used are mark down, write down, write on... maybe more)
https://hanabi.github.io/docs/beginner/critical-save-question-2
> Cathy would mark her slot 4 card as either:

I think being consistent in the wording would be good as well (if the wording is kept). e.g. change the ones that say "What identity does Bob write down?" to be "What identity does Bob write on the card?" because that is more direct.

Zamiell:
When you play Hanabi, you will **need** to write *card notes* and this page explains why.

<br />

## Explicit and Implicit Information

- In Hanabi, players will know *explicit* information and *implicit* information about the cards in their hand:
  - *Explicit* information is from the clues "on" the card.
  - *Implicit* information is from the context of the game *when* the card was clued.
- For example, in a 3-player game:
  - It is the first turn of the game and nothing is played on the stacks.
  - Alice clues Bob red, touching a card on his slot 1. Alice intends for this to be a *Play Clue*, signaling to Bob that he has a red 1 and that he should play it on the stacks right now.
  - From Bob's perspective, he knows that *explicitly*, his red card can be red 1, red 2, red 3, red 4, or red 5.
  - But Bob also knows that *implicitly*, Alice intends for him to play this red card, so it must be exactly the red 1.

<br />

## Players Need to Track Implicit Information

- On [Hanab Live](https://hanab.live/), the interface automatically will keep track of the clues that touch the cards. (And if you are playing Hanabi in real life, you could accomplish this by using counters.)
- However, this automated tracking of *explicit* information is not sufficient, because there is so much more *implicit* information than *explicit* information.
[...]

PR #720, opened 2021-07-20
Level 1 website features mentioned

vEnhance:
As promised in #713 

While I was writing it, occurred to me to mention empathy and rewind as well. I think {special notes, empathy, rewind} is approx the set of features that (a) is not obvious to beginners exists, and (b) is important to learn to use. But I don't feel strongly about this set.

Happy to have any wording in here tweaked (or rewritten completely) as needed .


Zamiell:
seems fine, thanks

(i edited it a bit and moved it to the bottom)

Issue #658, opened 2021-05-07
New hard variant convention: TCCM even if it gets 2 units of tempo


Lel0uch-H:
The [convention](https://hanabi.github.io/docs/level-6/#the-double-tempo-clue) states this:

> Typically, when a Tempo Clue is given that touches two or more cards, it means that all of the cards touched are playable in order from left to right.

In my experience, this is too rare to occur, especially for rank clues.

Suppose a game where not all the 1s have been played and the player has a bunch of 2s and a critical card right after that:

Played: r1, y1, b1

Bob's hand:  r4 g5 [2]r [2]g [2]b

In such cases, the rank 2 reclue conventionally says all of r2, g2 and b2 should play, which is so much more unlikely than only the left-most card is playable and the chop is important/critical.

I propose that we change this behaviour to a single tempo clue on the left most touched card and make it a tempo clue chop move.

pianoblook:
I'm not convinced that the *opportunity* to double tempo clue cards doesn't come up fairly regularly, it's just that people might not feel very excited to give the clue - there's a lot of potential greed in the idea that you can wait for the following connector to show up and just tempo them for free anyway. Or say for 2s, a lot of greed to just wait for all the 1s to get played and get to played for free.

Personally I think it's just an underutilized convention

Lel0uch-H:
The urgency to get a card played and cause a single chop move seems to be more than the urgency to play 2+ cards that lead nowhere. Hence, the proposal makes sense in my opinion.

Like you said, people might not feel very excited to give such a clue, because such a clue does not meet MCVP (gets no "new cards played). Getting a CM out of this instead seems to be better value both from a safety and efficiency perspective during the game.

sjdrodge:
This is a relatively rare situation (that a re-clue cannot be given that gives additional identifying information on the card), but TCCM is more useful than double tempo, so it makes a lot of sense to cater to that case. The current agreement strikes me as an artifact from the old old old days when people weren't even sure they wanted TCCM as a convention so a compromise was created.

Zamiell:
if you only want one card to play, focus the specific card.
if you want all cards to play, then focus all of them.
the *default* case is that either clue will not be blocked, the *special* case is that one of the clues is blocked.
the current convention configuration is (rightly) tuned for the default case.
[...]

PR #723, opened 2021-07-24
Implement #658

vEnhance:
I also broke off the list of convention changes into its own subsection since it was starting to get a bit on long side. Feel free to revert.


Zamiell:
nice

PR #735, opened 2021-08-20
Update level 1, 5 Stalls

Eliclax:
Update level 1 to say that 5 stalls in the early game can only be performed when there are no other available play or save clues.


Zamiell:
thx

PR #736, opened 2021-08-20
Pink Precedence & Turnabout Clarifications

pianoblook:
Added the section on Pink Precedence, as well as adding the same clarifications to Turnabout. 

Also realized the current Turnabout wording was stricter than intended; the 5 clue doesn't actually need to be given to exactly the 1 to work.


Lel0uch-H:
May need to clarify one more case: a one-away from chop 5 that causes an ejection can either be a BCME or a turnabout ejection. Can we define the precedence for that as well?

Lel0uch-H:
Oh nvm what I said, I missed that it is defined in another section.

PR #744, opened 2021-08-24
Added description for The Asymmetric Chop Move Dilemma

Lel0uch-H:


Zamiell:
if the trash bluff part necessary?
you can setup the example with a just a single 5 clue that is either a *5 Stall* or a *5's Chop Move*
and then say that any clue bob gives will be treated as just a 5 stall because no assync info

Lel0uch-H:
I think the example is necessary. The motivation for this issue came from a game where I was Bob in the trash bluff situation and I assumed necessary over unnecessary. The idea is that we should standardize both of these:
1. Bob should not be expected to figure out that a card has been asymmetrically chop moved.
2. Bob should be able to clue Cathy in an unambiguous manner without worrying about focus evaluation.

PR #746, opened 2021-08-24
Rename Shallow Discard/Misplay

jack67889:
When a teammate does an odd discard or even misplays, nothing related to ‘shallow’ seems to come to mind. By now I’ve already heard many times that people think it’s a nonsense name. Some time ago in a review, we had trouble discussing the situation, because people couldn’t think of the name of the convention. So we discussed better names and arrived at more descriptive ones:

Odd Discard
Even Misplay


Zamiell:
convention changes need to be posted as issues, not pull requests

PR #747, opened 2021-08-24
Rename playful ignition

jack67889:
Seems to be a pretty random and confusing name. Players are confused who should blind-play after calling it ‘ignition’. So I propose to make it clearer by changing the name to: Chop Move Pull (CM Pull).

(Probably the text should move to some other place in the doc after the rename.)


Zamiell:
convention changes need to be posted as issues, not pull requests

PR #750, opened 2021-09-02
Add a line about hyphen-ated group features

jack67889:
Beginners are always surprised when you tell them about this.


Zamiell:
DDA isn't a level 1 convention so i dont think this information is appropriate for level 1

jack67889:
Play order of 1s and sarcastic discards are level 3. This seemed to be the most appropriate section to me; I searched for 'website features' to find it myself. Or would you suggest adding an extra section in level 3?

Zamiell:
the main point of the hyphenated checkbox is for DDA alerting, so maybe a new section below where DDA is introduced

jack67889:
Okay, although I would think the ocm sound is most helpful when you still forget to pay attention to it.

Zamiell:
given that this pr hasn't been updated, i will close it

PR #757, opened 2021-11-10
Try to clarify what a Shout Discard is

DarthGandalf:


Zamiell:
can you put blaze discard stuff in a separate pr?

DarthGandalf:
Done.

To clarify the source of confusion: "This situation" refers to the situation in the previous sentence just before the word "this". But the latest situation in that sentence is Scream Discard. So it instead needs to refer to the situation before the previous one, which was still talking about discarding trash. The new text avoids this issue, and explicitly says what this new term means

Zamiell:
see my edits

Zamiell:
if that is okay with you then ill accept the pr

DarthGandalf:
> if that is okay with you then ill accept the pr

Such question implies that you'd wait for me to tell you whether it's okay. But LGTM

PR #765, opened 2021-12-02
Update pink.md

Dr-Kakashi:
Adding Taffy Double/Triple Bluff


SL765:
i3 is 2-away and i4 is 3-away right, so i5 should technically be a quadruple bluff, lol

PR #768, opened 2021-12-02
Move 5 stall's to level 2

### LINKED ISSUE #762: Move 5 stall’s to Level 2 (not level 1)
For newcoming players I think this is one less thing to worry about at first, and the *need* to 5-stall in No Variant at Level 1 is not that drastic.

vEnhance:
Closes #762


Zamiell:
there should be no mention of a 5 stall in level 1.
move the questions specifically relating to a 5 stall to level 2

vEnhance:
>  move the questions specifically relating to a 5 stall to level 2, if any

Do any of the later levels actually have questions right now? (Not sure where to put them.)

Zamiell:
> Do any of the later levels actually have questions right now? (Not sure where to put them.)

no
in this case, you will have to design new sections accordingly

Zamiell:
in the question where you deleted `The **only** legal action that Alice can perform is to clue number 5 to Donald as a *5 Stall* on the purple 5.`, we should regenerate that question in a slightly different form in the level 2 section

vEnhance:
Alright, shuffled around.

Zamiell:
fails lint

Zamiell:
thanks

PR #774, opened 2021-12-09
Remove the requirement for discarding the focus of trash bluff first

Lel0uch-H:


Zamiell:
you forgot to change the section on trash finesse

PR #775, opened 2021-12-09
Re-written to account for Always Loaded Principle

Lel0uch-H:


Zamiell:
for the white variant, does the *Always Loaded Principle* apply to:
1) save clues with number 2, number 3, and number 4
2) or save clues with number 3 and number 4
?

Zamiell:
if the former, why is 2 so special and 5 isn't

Lel0uch-H:
Because 2 needs to play so that 3 and 4 could follow, while 5 terminates the suit. To be consistent, we could add 5 saves in this umbrella as well, but I am sure it'll be very counter intuitive for even established players.

Zamiell:
> Because 2 needs to play so that 3 and 4 could follow

i dont understand this logic.
it seems perfectly consistent to me to play with a convention where only rank 3 and rank 4 save clues are *Loaded*.

Lel0uch-H:
More people could agree that getting w2 to play can be hard if it's the rightmost card with few more 2s in hand. Always loaded principle takes care of this case.

Btw, this is consistent with the original idea of white loaded play clues which included 2 saves as being loaded.

Zamiell:
ok

PR #813, opened 2022-01-13
Update save-clues.md

Cory847:
I don't think focus of a clue was previously discussed, should probably discuss it at some point.


Zamiell:
its not introduced until later.
all clues in this section only touch one card

PR #818, opened 2022-01-15
Fix Throw It in a Hole unusable clues formula

DrAnax:


Zamiell:
don't use prettier here please

Zamiell:
thx

PR #819, opened 2022-01-15
Add history of 5 pull

DarthGandalf:


Zamiell:
history doesn't belong on the website

DarthGandalf:
why not

Zamiell:
because it belongs in convention-reasons.md

PR #820, opened 2022-01-15
Move reasons to docs

DarthGandalf:


Zamiell:
i dont think it needs to be on the website, isn't it fine to have it in a separate document?

DarthGandalf:
a separate document which is hard to find?

don't you think the reason why the conventions exist is quite important for the readers of conventions? to better understand the logic behind them

DarthGandalf:
And the website allows the images, while the doc not so easily (without converting it to some other format first)

Zamiell:
> don't you think the reason why the conventions exist is quite important for the readers of conventions? 

no.
as a whole, the conventions are almost completely arbitrary.
the "reason" that any given convention exists is because it was voted in by an arbitrary amount of people at an arbitrary point in time, which is essentially meaningless.
the only important aspect is that you just have to memorize and/or know what the arbitrary agreement is

DarthGandalf:
Some of them are arbitrary, some of them are not. And the ones which are not arbitrary (which have some logic) are easier to memorize if you understand the logic

florian-5f3759df:
> > don't you think the reason why the conventions exist is quite important for the readers of conventions?
> 
> no. as a whole, the conventions are almost completely arbitrary. the "reason" that any given convention exists is because it was voted in by an arbitrary amount of people at an arbitrary point in time, which is essentially meaningless. the only important aspect is that you have to memorize what the convention is

I almost completely disagree with this way of phrasing the status of the conventions, and I can hardly believe you mean this as written. A statement like that is maybe a useful antidote against players claiming to play based on 'logic', but there is a long way between that and 'almost completely arbitrary'.

Besides the content-wise disagreement with what you said, I think statements like that also can put off new players from investing time trying to learn them.

Zamiell:
i dont want this information to show up in the search bar, so i think it is better the way it is for now

mangopie1:
Whenever I teach games, I use reasoning to explain a convention if it helps. 
For example, we want to save valuable 3s that is why 3 bluff exist. 
We want to save cards that wouldn't otherwise able to, that is why we usually use bluffs over finesses.
I haven't really look into the commits, but I think at least a link to the reasoning/history is good.

DarthGandalf:
There are benefits of having it visible described above.

Zamiel, what are benefits of NOT having it visible?

> i dont want this information to show up in the search bar,

That's why I put it at the very end of the menu

Zamiell:
whether or not it is in the menu doesn't affect whether it is in the search

DarthGandalf:
Ah, search. I see. Any way to skip certain pages from search?

DarthGandalf:
Is fixing https://github.com/algolia/docsearch-configs/blob/master/configs/hanabi-conventions.json#L10 enough?

PR #828, opened 2022-01-19
Update null.md

TheDaniMan:
A change to double finesses so that they make more sense and change of black 4 bluff to critical color bluff.


Zamiell:
thank you

PR #831, opened 2022-01-20
Rename the convention framework

DarthGandalf:
Because Hyphen-ated (the guy) explicitly asked to not attach his name to
it


BlueNovember:
This will inevitably lead to people asking "What does the H stand for?". 
Perhaps just rename it to "Hanab", as per the discord name?

pianoblook:
I've been intentionally calling it "H-Group" for awhile now, ever since I heard that Hyphen-ated explicitly asked us to not use his name.

lol it's pretty fucked up that we still do, but I've just assumed I held no power to do anything about it officially

NumerateClutter:
We should do this, and we should change the image (since the image is hyphenated’s profile picture)

“backronym”ing to hanab/hanab-live group seems pretty great to me — unambiguous and respectful of Hyphen-ated’s wishes

DarthGandalf:
@NumerateClutter any idea for the replacement picture?

Hyphen-ated:
Hi, thanks, it is correct that I would prefer this convention framework to not carry my name

DarthGandalf:
I replaced the image with pink 5 for now

florian-5f3759df:
I think it's very reasonable to follow the wishes of Hyphen-ated here, it's his name after all.

H-group is a very acceptable name to me, but we can also first make the decision that it has to go, and then start a suggestion+discussion (maybe +vote) procedure to pick something that will last.

Zamiell:
in the same way that my real life name of James does not belong to me and only me, the English word of "Hyphenated" does not belong to one guy who has decided to name himself "Hyphen-ated".

changing the name at this point is annoying and carries institutional friction.
i'm fine with changing "Hyphen-ated" to the more-netural "Hyphenated" and removing the prisoner logo and/or any other references to Hyphenated the person, but beyond that would be going too far

mmelwen:
Wait, I agree we should find a different name since Hyphen doesn't want us to use his but lets no be so hasty about it. Let's ask in about names and logos at discord first and choose the best one, we are using hyphen name for years I don't believe that he will object for 1-2 weeks to find something pretty to replace it

sjdrodge:
I'm sorry, why was this issue closed exactly? It seems to me that we have a large number of people saying it should be renamed and exactly one person saying that it should not. That person happens to have exclusive administrative access to this repo, but the claim is that decisions are made via consensus. Closing this issue under the circumstances makes clear the falsity of that claim, in my view.

Zamiell:
a random pull request is not the corrupt forum to discuss a name change, so i asked darth to open an issue

sjdrodge:
Fine.

Issue #825, opened 2022-01-16
Replace double bluffs with "don't blindplay after two consecutive blindplays"


TheDaniMan:
This is not a proposition for a convention change, but rather an equivalent reformulation of the current conventions.

Replace section 14 with the following: "if Alice gives a clue, and as a result of that clue both Bob and Cathy play a card in the following two turns, and Bob's card, Cathy's card and the clue do not all match each other, then the clue is considered a double bluff. No one should play any further cards from this clue, and the recipient of the clue should carefully go through every possibility that causes Bob to blindplay and Cathy to blindplay after that and write a corresponding note on their card. Note that 12 does not match 4, and 12/23 do not match 5".

The point is that one really doesn't have to know all the different types of double bluffs to give, play into, or receive them successfully. In practice, Alice just checks whether both Bob and Cathy will play from her clue, Bob and Cathy's front cards go brr, and the clue recipient checks all the possible cards rather than all the possible types of bluffs. I think this formulation is far clearer than 7 very similar examples.

Zamiell:
can you submit a PR so that i can see what it looks like

TheDaniMan:
Submitted, but I'd also add further examples if I had the tools to create the pretty pictures.

Zamiell:
i can show you how if you want on voice

DarthGandalf:
https://hanabi.github.io/docs/image-format should explain it

TheDaniMan:
I submitted a pull request for a new version of level 14 (which can still use a couple of edits + checking with other conventions that may now be accidentally covered) here: https://github.com/hanabi/hanabi.github.io/pull/837
There are two main points to this:
1) to simplify the conventions that need to be learned from "here is a list a bunch of conventions, each more identical than the last" to "stop playing after two cards get played", which I think is far more intuitive for newer players.
2) to give some examples on notes which new and sometimes not-so-new players quite often mess up, because the current double bluff section doesn't explain how the recipient of the clue should figure out what his card is (yes, you can deduce the procedure from various principles, but it's absolutely not the first thing a new player will think of after learning about double bluffs).

Zamiell:
i accepted the pr so i will close this for now, make another PR if you want to do any further updates

PR #837, opened 2022-01-22
Double Bluff Rework

TheDaniMan:
The formatting is weird in my preview, so please make extra sure I didn't royally mess up.
The changes are as follows:
1) a double bluff is defined as two people blindplaying in a row from the given clue (what was previously called "Double Bluff" was renamed to "Double Total Bluff"). The rest of the section is retained as examples.
2) 5 double bluff is replaced by 5-double-bluff-new, because it is not in line with the current treatment of 5 clues in low score phase.
3) an example of cluing a 2 with a 2 clue to get two 1s to play and an example of a color clue that matches the first blindplay in a rainbow variant are added, both emphasizing the note that Donald should put on his cards.
4) yml files are added in support of (2) and (3).

Note also that this makes pestilent double bluff redundant. The convention should probably be deleted in parallel, or the example also included in this section (relevant: pestilent triple bluff includes 'first, read pestilent double bluff'), but I didn't touch these in this patch.


Zamiell:
i thought the idea of #825 was that you would be deleting most of the text on this page and simplifying it, but that doesn't seem to be the case when looking at this PR

Zamiell:
look at what the website is now and do another PR if you want to change anything further

PR #839, opened 2022-01-23
Change outcome of ex 2 without relying on bluff targets

DarthGandalf:


Zamiell:
why does this pr have two separate changes

DarthGandalf:
Because it's easier this way, less merge conflicts with myself

Zamiell:
can you please put 1 change in 1 pr, that way it is easier to revert if needed

DarthGandalf:
Merge them via merge or rebase, without squashing them into one commit. Then it's easy to revert

PR #847, opened 2022-01-26
Learning path add trash finesse

mmelwen:


Zamiell:
its not supposed to list every convention, just the main ones

PR #882, opened 2022-02-15
Add interaction between Occam's Razor and Bob's Truth Principle

Lel0uch-H:


Lel0uch-H:
This has been the focus of many disagreements during reviews. Adding this here would be helpful I think.

Zamiell:
why is it at level 14 instead of e.g. after the Occams Razor section

Lel0uch-H:
Because before level 14, Bob cannot be in such a dilemma. Correct me if I am wrong.

Zamiell:
oh i see, because it has to do with double bluffs

Zamiell:
ok thanks, i did a few edits

PR #895, opened 2022-02-26
Remove PTT instructions

Lel0uch-H:


Zamiell:
why are we removing this

Lel0uch-H:
Because we do not have a PTT requirement for the H-group channels anymore.

Zamiell:
using PTT is recommended, but not enforced, thus we should keep the instructions as-is

PR #899, opened 2022-03-17
Fix a typo

crossproduct42:
Changed "move information" to "more information" in the solution.


Zamiell:
this is the same pr as the other one???

PR #915, opened 2022-04-19
Fix "clue giver" tag in Save Principle (Question 1)

nickm2402:
Question page: https://hanabi.github.io/docs/beginner/save-principle-question-1

Bob is tagged as clue giver. The question starts with "Alice clues blue to Bob..." and it is now Bobs turn, so Alice should get the clue giver tag. 

This probably happened because the solution is, that Bob needs to clue now (also the tag spoils the answer in a way).

Not sure if I got that yml format correct. Please check before merging


Zamiell:
i guess there doesn't need to be a clue giver tag for this question

PR #919, opened 2022-04-26
Remove extra player in pestilent-double-bluff.yml

vEnhance:
The text accompanying this figure references a four-player game, but the image was a five-player game.


Zamiell:
thanks sir

PR #923, opened 2022-05-22
Level 19: Clarify and add an example

argothiel:
Clarify Occupied Play Clues and Occupied Finesse.
Add an example of an Occupied Finesse.


Zamiell:
thanks

PR #928, opened 2022-05-29
Change some texts to fix error or make it clearer.

jk15162428:
In rainbow-ones part, the green and blue doesn't match, so change one.

In 4-charm part, "trash" is not precise, cuz if that card is b3, b4, etc, which are not playable, not trash.


Zamiell:
thanks

PR #931, opened 2022-05-30
Add question for reverse finesses to level 2

GuolinM:


GuolinM:
Demo images:
image
image

Zamiell:
no need for demo images, i'll check out the branch

GuolinM:
I plan to add more later to level 2 but wanted to submit this first to see if there's any concerns on how I'm formatting these

Zamiell:
thanks for adding.

see my edits here: https://github.com/hanabi/hanabi.github.io/pull/931/commits/c6a6f320d9992a38de592a4c445ad4368109b5ea

also, these questions are kind of basic, you should add one where the right answer is not to do the obvious thing.

for example, notice how here:
https://hanabi.github.io/docs/beginner/5-save-question-1
the obvious answer for the first question about a 5 save is to do a 5 save.
but cleverly, the right answer is to not 5 save.

GuolinM:
> thanks for adding.
> 
> see my edits here: [c6a6f32](https://github.com/hanabi/hanabi.github.io/commit/c6a6f320d9992a38de592a4c445ad4368109b5ea)
> 
> also, these questions are kind of basic, you should add one where the right answer is not to do the obvious thing.
> 
> for example, notice how here: https://hanabi.github.io/docs/beginner/5-save-question-1 the obvious answer for the first question about a 5 save is to do a 5 save. but cleverly, the right answer is to not 5 save.

Yup, it would be great to add some "trick" questions. But I wasn't sure how many questions we can fit at the bottom of the document - would you be fine with adding potentially 10 questions in total? (2 or so questions per new convention in the level) Or shoudl we find some other organization system

Zamiell:
> would you be fine with adding potentially 10 questions in total?

yes, you can add as many as you like.

split it off into a separate page if you think that it clutters the page too much to have too many questions.

PR #933, opened 2022-06-01
Overhaul how challenge questions are laid out in the website

GuolinM:
Creating this in anticipation of more questions to come in each section. I'm not sure about your thoughts on this layout so feel free to change as needed


Zamiell:
seems fine

package.json
```diff
@@ -39,6 +39,7 @@
     "isaacscript-lint": "^1.1.6",
     "isaacscript-tsconfig": "^1.1.17",
     "npm-check-updates": "^13.0.3",
+    "prettier": "2.6.2",
```
Zamiell, line 42: its already included in isaacscript-lint so you don't have to put it here

PR #935, opened 2022-06-02
Reverse finesse and self-finesse quizzes

GuolinM:
Lots of quizzes in this one, but should have good coverage of the reverse finesse and self-finesse conventions. The other level 2 conventions will still need to be added at some point


Zamiell:
this pr doesn't contain #932 so you probably need to rebase it on main

Zamiell:
see my edits, thanks

GuolinM:
> see my edits, thanks

Thanks for the edits!

PR #960, opened 2022-08-01
Update trash-touch-elimination.yml

Romain672:
Double check please but on this picture: https://hanabi.github.io/docs/level-16#trash-touch-elimination I don't believe the b1/g1/y1 part is important.
Clueing 4 here usually is enough fine in a normal game.
So clueing red, except in rare cases should give elimination notes on r5 whatever cards Bob is holding.


Zamiell:
yes you are right

Zamiell:
i think its better for the cards to just be fully displayed, since the situation is from alice's perspective

Zamiell:
i edited your pr accordingly

PR #961, opened 2022-08-02
Level 18: add unknown cards first

Romain672:
Why not put unknown cards second to have one less section on the document? That resolve automatiquely the below section.
If it's annoying you could put two 1 priority.

### DIFF
--- a/docs/level-18.mdx
+++ b/docs/level-18.mdx
@@ -24,17 +24,17 @@ import TrustFinesse from "@site/image-generator/yml/level-18/trust-finesse.yml";
 
 - Typically, players will only have one card to play at a time. In this case, unless there is some really good clue to give, there are no decisions to be made - they just play their one playable card.
 - What if a player has two or more playable cards to play? Which card should they play first?
-- If one of the cards is not completely known yet (e.g. a playable 2 of unknown color), then the player might want to play that card first in order to find out exactly what it is. In general, **nothing special is triggered by players playing an unknown card.**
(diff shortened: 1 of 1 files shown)


GuolinM:
I think this doesn't work because, if I understand correctly, nothing special happens if a known card is played over an unknown card if *not all* possibilities of the unknown card have higher priority than the known card.

Imagine if all stacks were played up to 3, except red, which doesn't have any cards yet. Alice has [r1] x x p[4]. Bob has a p[5] in his hand somewhere. If Alice plays r1, it shouldn't trigger a prio - she doesn't know for sure if her 4 leads into Bob's p5.

Romain672:
Uh, I tried to re-read the section and I'm lost now :(

iamwhoiamhahaha:
Romain, you should probably open an issue for that rather than opening a pull request. 

Also the table should be like this as Guolin has mentioned:
```
| 2 | Cards that lead into clued cards in someone else's hand | Otherwise, the team would lose _Tempo_.             |
| 3 | Cards that lead into the player's own hand              | It is bad for a suit to be "held up" on one player. |
| 4 | 5's                                                     | Playing a 5 gets the team a free clue.              |
| 5 | The lower rank card                                     | The smaller stacks are more important to fill up.   |
| 6 | Unknown card                                            | Gaining information about his own hand is valuable. |
| 7 | The left-most card                                      | The left-most card is more likely to be good.       |
```

Romain672:
> Romain, you should probably open an issue for that rather than opening a pull request.

I was expecting a short no in fact. But if it's more complicated than that, I will open one.

Zamiell:
ok, i updated the table to what jeff said

Zamiell:
sounds like we need to close this PR then

PR #962, opened 2022-08-02
The Out-of-Order Corollary

Romain672:
If someone has r4-r3-x-x-x with r3 playable, and he receives a red clue, it's still possible that it is a bluff, if the 3 clue is available and clear.


Zamiell:
that's an edge case that we don't care about documenting though

PR #971, opened 2022-08-11
Update level-2.mdx

argothiel:


Zamiell:
looks like your edit has an error in it

argothiel:
What error?

Zamiell:
i mean just look at it, you are combining a parenthetical with a half-sentence

argothiel:
Thanks, I've updated the fix. Next time, please try and be more specific with your feedback. If I missed the error once, your "just look at it" isn't very helpful.

Zamiell:
looks good thanks

PR #973, opened 2022-08-11
Update unnecessary-trash-finesse-trash-push.yml

Romain672:
Missed the 'border false' first time i was editing those examples :(


Zamiell:
please test your prs locally before submitting

PR #612, opened 2021-03-18
BCME and Playful Ignitions

pianoblook:
Not sure why all the Dark Pink stuff is also showing up as 'changes', since those have already been added.

PR #974, opened 2022-08-11
Remove 5 Double Pull examples [#612]

argothiel:
The 5 Double Pull convention is obsolete due to #612. This commit
removes the leftover 5 Double Pull examples.


Zamiell:
good thanks

PR #977, opened 2022-08-17
Add link to VARIANTS.md

vEnhance:
Feel free to edit the wording however you feel like, I just made something up.


vEnhance:
Incidentally, I should mention that 2-player and empty clues don't really feel like they belong on this page, but I dunno if there's a better place to put them.

Zamiell:
seems fine, thanks

PR #981, opened 2022-08-18
Clarified definition of trash in BCME

omegaxisJC:
Added a clarification that chop moving duplicates should be allowed and should not call for a BCME when it can be for distribution purposes.


Zamiell:
seems relatively uncontroversial so i will merge it, thanks

PR #988, opened 2022-08-27
Update good-touch-principle-question-1.yml

mkillianey:
According to the solution, Alice _needs_ to clue the b3:

https://github.com/hanabi/hanabi.github.io/blob/c3d0bc4c77ff771e32a4c78fb884139795cdaef6/docs/beginner/good-touch-principle-question-1.mdx?plain=1#L28

But this seems to be making a lot of assumptions about what else is happening in the game.

Adding the b3 to the discard pile so that she genuinely _needs_ to clue it.

### DIFF
--- a/image-generator/yml/beginner/good-touch-principle-question-1.yml
+++ b/image-generator/yml/beginner/good-touch-principle-question-1.yml
@@ -4,6 +4,8 @@ stacks:
   - g: 5
   - b: 2
   - p: 1
+discarded:
+  - b3
 players:
   - clue_giver: true
     cards:


mkillianey:
Hmm...although now I see that it specifically says _Play Clue_ and not _Save Clue_.

I think the point here is that the question asks what Alice should do, and it might be reasonable for Alice to discard, too. 
 Maybe the question could be "If Alice wants to give a clue, which clue should she give?"

Zamiell:
> this seems to be making a lot of assumptions about what else is happening in the game.

no?
what assumptions do you think it is making

Zamiell:
> Adding the b3 to the discard pile so that she genuinely needs to clue it.

whether or not the b3 is in the discard pile is irrelevant as to whether Alice "genuinely needs to clue it".
in both scenarios, she genuinely needs to clue it.

mkillianey:
As someone new-ish to Hanabi and brand new to your conventions, I was going through the beginners guide and hadn't encountered anything that said "If you can clue a playable card, you must do so."  (I think that is discussed in the _next_ section, the Save Principle, but the Good Touch Principle comes before that in the beginner's guide, so that hasn't been covered yet.)

Observation: if the question was phrased "What clue should Alice give to Bob?" would have clued me in that discarding is a non-option, and the challenge is to decide which card to clue, and whether to give color or value.

Zamiell:
>  "If you can clue a playable card, you must do so." (I think that is discussed in the next section, the Save Principle, but the Good Touch Principle comes before that in the beginner's guide, so that hasn't been covered yet.)

i see.

> Observation: if the question was phrased "What clue should Alice give to Bob?" would have clued me in that discarding is a non-option, and the challenge is to decide which card to clue, and whether to give color or value.

i agree, can you submit a PR for that?

PR #989, opened 2022-08-29
Minor clarification in 2 save discussion

alercah:
Once one of two 2s are discarded, the other can be Critical Saved, not just 2 Saved. But a functional change but I think it helps foreshadow that 2s can be saved by colour once they are critical.


Zamiell:
you are right

Zamiell:
actually this is wrong, i deliberately put 2 save because the term "Critical Save" has not yet been introduced at this point in the docs

alercah:
It's introduced at the bottom of
https://hanabi.github.io/docs/beginner/save-clues

Zamiell:
i dont think that the word appearing in a table of contents counts for the purposes of this

alercah:
Personally I think it is fine since it was very recent, people are expected
to read this section in order, and the name is pretty self explanatory
anyway. But it's hardly a big deal.

PR #990, opened 2022-08-31
Summary page

Romain672:
Here is the base of the text, still neeed:
 - create the page, the links between page, at the right place on the document (atm, it's just a page which is not used) 
 - put the titles in bold (weird that they are in bold in others levels but you can't use the * )
 - maybe add extra files at the start of the page, like on others levels pages

And:
 - separate the document in more sections (like finesses can englobe 4 sections)
 - double check some things

### DIFF
--- a/docs/summary
+++ b/docs/summary
@@ -1,0 +1,257 @@
+---
+id: summary
+title: Summary
+---
+
+## Beginner
+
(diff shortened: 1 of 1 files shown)


Zamiell:
this doesn't pass ci

PR #996, opened 2022-09-23
Incorrect clue recipient listed in example

Kylewolfson:
Updated the last bullet of Example 2 of "Cluing 1's in the early game" to the correct clue recipient for Alice (was written Donald, should be Cathy)

Unless this was a meta-joke about using context to interpret the given text. =)


Zamiell:
it was not a meta-joke, thanks

Zamiell:
although the doc could probably use some more humor

PR #997, opened 2022-09-27
fix

variussandry:
The question says “all remaining useful cards are gotten”, so likely it means g5 has been played.


Zamiell:
thx

PR #1012, opened 2022-10-22
Fix Ex 3 of Unnecessary Moves with Unknown Trash Off Chop

chennisden:
The example has b2 play when b2 is already played on the stacks. Fixes by making b2 not yet played on the stack.


Zamiell:
thanks

PR #1014, opened 2022-10-28
BGA -> H-Group

fairyboots:
Making an inclusive environment for those transitioning.


Zamiell:
just needs the levels to be updated and i will merge

PR #1017, opened 2022-10-29
Bga -> H-Group

fairyboots:
Bro where is this pull request even going... edited the edit, looks good! less floral language but I like the brevity a lot. Not sure what other name we can come up with


fairyboots:
oops pls don't discharge from this dupe

PR #1020, opened 2022-10-31
Update level-8 stall chart

newduke:


Zamiell:
putting chop moves in this section would imply that they are mandatory.
more specifically, it would imply that because you did not perform some chop move, it would invoke a priority move.

Zamiell:
i will close this for now

PR #1025, opened 2022-11-07
Update learning-path.mdx

heff10:
Following changes made to the list of levels in the doc, this will also change the rough recommendations to how the simple variants map to convention level minimums. The exact numbers aren't important, but every time the doc is changed, these numbers should also change.


Zamiell:
sounds good thanks

PR #1041, opened 2022-11-27
Update level-19.mdx

pianoblook:
We'll still need to make new example images for 5NE & 5ND.


Zamiell:
Can you do that as part of the PR?

pianoblook:
I promise I would have if I knew how, sorry

vEnhance:
Added the requested images.

Zamiell:
thanks

PR #1048, opened 2022-12-18
Update prism.mdx

argothiel:
This pull request merges together two points which are talking about the same case.


Zamiell:
thanks

PR #1051, opened 2022-12-23
Update level-22.mdx

variussandry:
Of course it could be blue, it *is* blue. The point the doc is trying to make here is that, just because it's not known playable, doesn't mean it *cannot* be playable.


Zamiell:
ok

PR #1052, opened 2023-01-02
Update level-19.mdx

jk15162428:
I'm quite sure this is a mistake, but not to be 100% sure if that's what it should be.


Zamiell:
Thx

PR #1055, opened 2023-01-16
Update level-20.mdx

ReaverSe3:
i.e 5, black, red in muddy also can be Load Clue


NZadeh (changes requested):
nit on the wording/formatting: "A _Load Clue_ applies even if the clue would look like a save clue otherwise."

Zamiell:
thanks

Issue #1045, opened 2022-12-04
Typo / misunderstanding under Taffy Double Bluff? Rename it? 😄 


NZadeh:
**Edited**: 

Taffy and Bubblegum are the same. I attempt to convince you of that in the linked PR. Thanks for your time!

-----

**Original**:

Section: https://hanabi.github.io/docs/variant-specific/pink#the-taffy-double-bluff

> Rarely, a player can perform a Taffy Double Bluff by violating Pink Promise on a three-away-from-playable pink card.

The sentences further below actually give examples and explicitly state using "two-away-from-playable", so I think this first one is an outlier/typo.

Furthermore, I don't think the "X-away-from-playability" is the key logical insight here. In my mind, the logic for this move is more akin to the Pestilent Double Bluff, in that Donald would otherwise incorrectly mark their card after the first blind play resolves (as if it were a Bubblegum Bluff).

The thing that makes the Taffy Double Bluff play is that the Bubblegum Bluff would be a lie.
(In contrast, the thing that makes a Taffy [Single] Bluff known is a Bubblegum Bluff and conventional Bluff are impossible.)

~~In fact, I might motion to change the naming of this bluff to something like Pestilent Bubblegum Bluff... :)~~

~~This would naturally extend to situations where e.g. an i3 is clued in Donald's hand with 4, getting Bob to play i1 from slot 4, and forcing Cathy to blind play to "resynchronize" the information (showing that neither Donald nor Bob has the i2 in slot 4 or slot 1 respectively)~~   _EDIT: nevermind, this move is illegal. The docs promise a Finesse._

Zamiell:
can you submit a pr

NZadeh:
Done, please take a look!

Zamiell:
looks like its not done

NZadeh:
Okay, we've got [a winner](https://github.com/hanabi/hanabi.github.io/pull/1060). Now you can take a look for real :)

NZadeh:
also who could have predicted that
> can you submit a pr

would turn into an 8 hour fun-fest with doc reviews and git-wrangling...? 😢 😭

NZadeh:
@Zamiell please let me know if you have any questions or feedback!

NZadeh:
Beep beep :) @Zamiell

Zamiell:
i'll take a look when i get an hour or so of spare time

NZadeh:
[...]

PR #1057, opened 2023-01-18
Issue #1045: Rename the Taffy Double Bluff, and update the language in Taffy Bluff

NZadeh:
The logic behind a Taffy Double Bluff is much closer to that behind a Bubblegum Finesse or a Pestilent Double Bluff. We are renaming it as such.

The description for Taffy Bluff was not logically motivating and, as such, could be hard for first-time readers to understand. (I know I struggled with understanding it...) I'm hoping the new rewording makes it more intuitive.


NZadeh:
I'm not convinced I did this right, so guidance would be welcome. Wordsmithing also welcome.

This is addressing https://github.com/hanabi/hanabi.github.io/issues/1045. There's probably a structured way of linking them, but I'm also not sure how to do that. Thanks!

NZadeh:
Sorry, I'm fumbling my way around git. I was hoping there was a way to collapse ("squash"?) all of these commits into one... as I made a mistake in the first one. Hopefully a reviewer can do that for me in the UI?  https://stackoverflow.com/a/43858707

But anyways, the "Files" view should have the "sum" of all the commits anyways.

Zamiell:
prs are squashed as they are merged automatically via the github ui, so you dont have to manually worry about the number of the commits in the pr

PR #1058, opened 2023-01-18
Rename the Taffy Double Bluff, and update the language in Taffy Bluff [[Issue #1045]]

NZadeh:
The logic behind a Taffy Double Bluff is much closer to that behind a Bubblegum Finesse or a Pestilent Double Bluff. We are renaming it as such.

The description for Taffy Bluff could benefit from logical motivation, without which it could be hard for first-time readers to understand. (I know I struggled with understanding it...) I'm hoping the new rewording makes it more intuitive.

I'm not convinced I did this right, so guidance would be welcome. Wordsmithing also welcome.

This is addressing https://github.com/hanabi/hanabi.github.io/issues/1045. There's probably a structured way of linking them, but I'm also not sure how to do that. Thanks!


NZadeh:
Sorry, I think there are still issues with the rewrite that need working out.

For example, it is possible to give Bubblegum-like clues on two-away-from-playables that require only 1 blind play to synchronize or might require 2 blind plays. I'll be back.

PR #1060, opened 2023-01-19
Unify Bubblegum and Taffy moves. Improve flow of Bubblegum section. [[Issue #1045]]

NZadeh:
After much thought and deliberation (also thanks to HelanaAshryvr for walking through this with me), it has become clear that Bubblegum and Taffy are actually the same.

~~The difference, if any, is whether the number the pink card could be is actually greater than (Taffy) or less than (Bubblegum) the clue's number.~~ _EDIT: this is not quite accurate._

I also tried to improve the narrative of the section. Thoughts and suggestions for improvements are, of course, welcome.


NZadeh:
@Zamiell PTAL

docs/variant-specific/pink.mdx
```diff
@@ -253,7 +256,24 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 
 <BubblegumFinesse />
 
-- _Bubblegum Finesses_ can only be used to get pink cards.
```
Zamiell, line 256: why is this line deleted

docs/variant-specific/pink.mdx
```diff
@@ -253,7 +256,24 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
+- Whenever a pink card plays in response to a _Bubblegum_ move, it is guaranteed to the team to be a _Finesse_.
```
Zamiell, line 259: i dont understand this line.
isn't it false because bubblegum bluffs exists?

docs/variant-specific/pink.mdx
NZadeh: it's replaced with the line below it. I thought they were identical, with the second one being clearer to someone who is learning. 

given your comment on that line as well, it seems that may not be the case?

how do you interpret this (deleted) line?
"Bubblegum Finesses can only be used to get pink cards."

what is it actually saying?
I think the only thing it _can_ be saying is that once a pink card plays, it's a Finesse.

docs/variant-specific/pink.mdx
NZadeh: as far as I understood, there is no such thing as bubblegum-bluffing a pink (or omni) card. those are finesses. (hence the reword above.)

similar to Bob playing red 1 into a red clue on Cathy's red 3. it's not a bluff. either Cathy bombs red 3 or Bob bombs slot 2 next round.

docs/variant-specific/pink.mdx
```diff
@@ -253,25 +256,69 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
 
+### The Bubblegum Bluff (Two-Away-From-Playable Form)
+
+- Normally, _Bubblegum Bluffs_ are always performed by touching _one-way-from-playable_ pink cards.
+- Rarely, it is possible to perform a _Bubblegum Bluff_ on a _two-away-from-playable_ pink card. However, this can only be done in a specific situation where, due to the circumstances of the clue, it would be impossible for the clued card to be _one-way-from-playable_.
+- For example, in a 3-player game:
+  - It is the first turn and nothing is played on the stacks.
+  - Alice clues number 2 to Cathy, touching a pink 3 on slot 3 as a _Play Clue_. (The pink 3 is currently _two-away-from-playable_.)
+  - Bob knows that Alice is violating _Pink Promise_.
+  - If Bob blind-plays his _Finesse Position_ card and it successfully plays, it is guaraunteed to be one of the 1s. If that happened, then Cathy would mark her clued card as the matching 2 and will go on to misplay it.
```
NZadeh, line 301: (oops, I need to fix this typo... will do that sometime before submission.)

docs/variant-specific/pink.mdx
```diff
@@ -253,25 +256,69 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 
 <BubblegumFinesse />
 
-- _Bubblegum Finesses_ can only be used to get pink cards.
+- Whenever a pink card plays in response to a _Bubblegum_ move, it is guaranteed to the team to be a _Finesse_.
+
+### The Bubblegum Prompt
+
+- Normally, you can only _Prompt_ a card if it is the left-most clued card. However, you can _Prompt_ a non-left-most clued card by violating _Pink Promise_. This is called a _Bubblegum Prompt_, which conceptually similar to a _Bubblegum Finesse_.
```
NZadeh, line 261: s/which/which is

docs/variant-specific/pink.mdx
Zamiell: what is the typo?

NZadeh: some check was failing: "guaranteed" only has one "u"

Zamiell:
ok, take another look at the pr to see if i made any mistakes

docs/variant-specific/pink.mdx
```diff
@@ -291,37 +291,33 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
-- The clue-giver must be expecting additional blind-plays to resynchronize the information.
+- But what if the note that the clue receiver would write on their card is still incorrect? In this situation, the clue-giver expects additional blind-plays to resynchronize the information, in the same way that a normal _Double Bluff_ works.
 - For example, in a 4-player game:
   - It is the first turn and nothing is played on the stacks.
   - Alice clues number 3 to Donald, touching a pink 4 on slot 1.
-  - Bob knows that Alice is violating _Pink Promise_. This must mean that he is supposed to play his slot 3 card as pink 1, with pink 2 and 3 promised as playable through his current finesse position.
-  - Bob blind-plays slot 3, and it is a red 1.
-  - From Donald's perspective, he knows that Alice performed a _Bubblegum Bluff_. Donald marks his slot 1 card as a pink 2 (_one-away-from-playable_).
+  - Bob knows that Alice is violating _Pink Promise_. This must mean that he is supposed to play his slot 3 card as pink 1, with pink 2 and 3 promised as playable through his current _Finesse Position_ and _Second Finesse Position_, respectively.
```
NZadeh, line 313: (d'oh, I just realized you have made edits to the PR...)

I think my original intention was to say that this is a finesse, hence the word "through". it could be layered, etc. 

not sure if the new language about "Second Finesse Position" and "respectively" is allowing for that? it's probably "clear enough" ? but maybe worth being extra clear by not mentioning it.

docs/variant-specific/pink.mdx
```diff
@@ -291,37 +291,33 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 - _Bubblegum Bluffs_ can only be performed on non-pink cards; otherwise, it will look like a _Bubblegum Finesse_.
 - Note that it is impossible to _3 Bluff_ and _Bubblegum Bluff_ at the same time.
 
-### The Bubblegum Bluff, part 2
+### The Bubblegum Bluff (Two-Away-From-Playable Form)
 
-- Note that, as with a standard _Bluff_, the receiver of a _[Bubblegum Bluff](#the-bubblegum-bluff)_ marks their pink card as the lowest-possible logical _Good Touch_ value it could be.
-- However, it is sometimes possible that the lowest-possible _Good Touch_ value is actually _two-away-from-playable_.
+- Normally, _Bubblegum Bluffs_ are always performed by touching _one-way-from-playable_ pink cards.
+- Rarely, it is possible to perform a _Bubblegum Bluff_ on a _two-away-from-playable_ pink card. However, this can only be done in a specific situation where, due to the circumstances of the clue, it would be impossible for the clued card to be _one-way-from-playable_.
```
NZadeh, line 296: @Zamiell obviously, you thought my proposed wording was unclear, hence the changes 😛 

but I sort of disagree with this wording (which reminds me more of the pre-PR document).
it's not really a "specific situation" or "circumstance" -- it's precisely the _logic_ following from _Good Touch_ and lowest-value.

what were your thoughts on this?

docs/variant-specific/pink.mdx
NZadeh: hmm... did we accidentally remove the other statement?

I think it's important to communicate somewhere that Bubblegum moves causing pink cards to blind play are finesses.  That might now be missing from the doc? or is it elsewhere?

NZadeh: (oh, we do have "- _Bubblegum Bluffs_ can only be performed on non-pink cards; otherwise, it will look like a _Bubblegum Finesse_." under Bubblegum Bluffs... but hmmm, is that clear enough?)

NZadeh:
thanks for taking a look @Zamiell!

I left some comments... if you want a "typo-review", I can do that once any other edits (in response to outstanding comments) are done or declined. 😄

docs/variant-specific/pink.mdx
Zamiell: > it could be layered

yeah but all finesses can be layered, so I don't think we need to specifically mention that.

docs/variant-specific/pink.mdx
Zamiell: does it have to do with "good touch" though?
like, say for example you do a "2 save" and it causes a crazy blind play, then it can't be a 2 save anymore, and it is a taffy bluff as a 3.
in that example, the operant principle is the fact that the "2 save" convention already exists and that since it exists, it wouldn't cause any blind plays.
but the doc can't be worded like this:
> Sometimes, a clue would look like a _2 Save_, but it can't be a _2 Save_, so then it would have to be on the next _two-away-from-playable_ card.
so we need to word that sentence in a more generic way, so i chose the phase "circumstances of the clue", but you can workshop this if you don't like it.

NZadeh: ah okay I see what you're saying.

the "logic" relies on other conventions as well, like 2 saves existing.

but in a way, that's the point. this convention is built on top of those existing ones. we know that a 2 save doesn't cause blind plays. so if there is a blind play, the card can't be a 2.

and we agree that lying about a card's identity needs to be fixed in a certain way (blind play from number slot).

given those facts, the card being one away or two away isn't really important -- it's the lowest possible value it could be, assuming the card is useful ("good touch").

That's precisely where (lowest possible and) good touch comes in: "what value is the card?"

docs/variant-specific/pink.mdx
Zamiell: i suspect so

docs/variant-specific/pink.mdx
Zamiell: I updated the doc a bit if you want to review the pr again and suggest more edits (maybe we should use the GitHub suggest edit feature, if you're familiar with that)

NZadeh: sorry been traveling, on mobile right now, I can take a look and try to suggest something

docs/variant-specific/pink.mdx
```diff
@@ -253,24 +256,67 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
+
+- Normally, you can only _Prompt_ a card if it is the left-most clued card. However, you can _Prompt_ a non-left-most clued card by violating _Pink Promise_. This is called a _Bubblegum Prompt_, which is conceptually similar to a _Bubblegum Finesse_.
+- For example, in a 4-player game:
+  - All the 1's are played on the stacks.
+  - Bob's _Finesse Position_ card has negative pink. He also has three cards clued with number 5 on slot 2, slot 3, and slot 4.
+  - Alice clues number 3 to Donald, which focuses a pink 4 on slot 2 and touches a pink 3 on slot 3.
+  - Bob sees that in order to make the pink 3 and the pink 4 playable, someone has to have the pink 2. Since no-one else seems to have it, Bob must have it as one of his clued 5's.
+  - Given that Alice violated _Pink Promise_, Bob can determine that his pink 2 is on his slot 3. Bob plays his slot 3, and it is the pink 2.
+  - Cathy discards.
+  - Donald knows that normally, clues do not cause non-left-most cards to become _Prompted_. Thus, Donald knows that Bob must be responding to Alice's _Bubblegum Prompt_ on a card that is not actually a 3.
```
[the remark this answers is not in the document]
NZadeh: actually, I'd probably reword the sentence to something like:

"Donald knows that clues do not normally _Prompt_ a non-leftmost card."

Or drop the double negative, but it loses a bit of it's intention.

docs/variant-specific/pink.mdx
```diff
@@ -253,24 +256,67 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
+  - The next card after a pink 3 is a pink 4, so Donald knows that the card that Alice focused is a pink 4. Subsequently, since Alice's clue was a _Play Clue_, the pink 4 must be playable. In order for a pink 4 to be playable, someone has to have a pink 3. Donald does not see it anywhere else, so he determines that he must have the pink 3 on slot 3 (since that is the only other clued card that could possibly be a pink 3).
```
NZadeh, line 270: I'm trying to workshop ways to simplify/condense the statements. The first sentence here sounds a bit weird as it's an obvious statement, but it's not clear how it immediately relates to the bubblegum logic.

"Thus, Donald knows that the focus of Alice's _Play Clue_ cannot be a 3; it must be a pink 4. In order for the 4 to be playable, someone must have a pink 3. ..."

docs/variant-specific/pink.mdx
NZadeh: My point was that the language "through his _Finesse Position_" should be sufficient, without mentioning second finesse, since that may not even be where the card is located. I.e. the current language is more specific and hence inaccurate in some cases.

docs/variant-specific/pink.mdx
```diff
@@ -253,24 +256,67 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
 <BubblegumBluff />
 
-- _Bubblegum Bluffs_ can only be performed on non-pink cards; otherwise, it will look like a _Double Finesse_ with one part _Bubblegum Finesse_ and one part normal _Finesse_.
+- _Bubblegum Bluffs_ can only be performed on non-pink cards; otherwise, it will look like a _Bubblegum Finesse_.
 - Note that it is impossible to _3 Bluff_ and _Bubblegum Bluff_ at the same time.
+- Note that whenever a pink card plays in response to a _Bubblegum_ move, it is guaranteed to the team to be a _Finesse_. (In other words, you cannot use a _Bubblegum Bluff_ with a target of a pink 3 to bluff out a pink 1.)
+
+### The Bubblegum Bluff (Two-Away-From-Playable Form)
+
+- Normally, _Bubblegum Bluffs_ are always performed by touching _one-way-from-playable_ pink cards.
```
NZadeh, line 295: Sorry, I'm on mobile, so I can't easily find the old language. I really do prefer it though, as I had reviewed that multiple times already before submitting it for review.

Can you think of a less invasive way to update the old language if you feel it needs additional emphasis on the fact that other conventions like 2 saves exist?

NZadeh (commented):
Accidentally entered this review mode, so here you go :)

docs/variant-specific/pink.mdx
Zamiell: not sure, suggest something using the suggest feature

NZadeh:
```
- Note that, as with a standard _Bluff_, the receiver of a _[Bubblegum Bluff](#the-bubblegum-bluff)_ marks their pink card as the lowest-possible, useful value it could be (given the team's preceding actions).
- However, it is sometimes possible that the lowest-possible, useful value is actually _two-away-from-playable_.
```

What about something like this? (I'll find the suggest feature shortly).

- Got rid of "good touch" and replaced it with "useful"
- Added "(given the team's preceding actions)" to hint at the fact that there might be blind plays, etc... a 2 save does not a bluff make :)

NZadeh, line 294:
```suggestion
- Note that, as with a standard _Bluff_, the receiver of a _[Bubblegum Bluff](#the-bubblegum-bluff)_ marks their pink card as the lowest-possible, useful value it could be (given the team's preceding actions).
```

docs/variant-specific/pink.mdx
```diff
@@ -253,24 +256,67 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
...
+- Rarely, it is possible to perform a _Bubblegum Bluff_ on a _two-away-from-playable_ pink card. However, this can only be done in a specific situation where if the clued card to be _one-way-from-playable_, it would not work out.
```
NZadeh, line 295:
```suggestion
- However, it is sometimes possible that the lowest-possible, useful value is actually _two-away-from-playable_.
```

docs/variant-specific/pink.mdx
NZadeh, line 294:
```suggestion
- Note that, as with a standard _Bluff_, the receiver of a _[Bubblegum Bluff](#the-bubblegum-bluff)_ marks their pink card as the lowest-possible, useful value it could be (given the team's preceding actions).
```

NZadeh (commented):
Tried to add changes and keep them in a single batch... let's see if it works.

docs/variant-specific/pink.mdx
NZadeh: Could also drop the leading "Note that, ..." and start with "As with a standard _Bluff_, ...."

NZadeh: (oops, dupe suggestion? internet lag I suppose)

docs/variant-specific/pink.mdx
NZadeh: also wouldn't be opposed to also wording it like:

"However, it sometimes happens that the lowest-possible, ...."

NZadeh: what I want to emphasize is that this not really a "rare" thing. it's just a "numbers + logic" thing depending on what clue is available for what card with what on the stacks.

docs/variant-specific/pink.mdx
Zamiell: > Note that, as with a standard _Bluff_, the receiver of a _[Bubblegum Bluff](#the-bubblegum-bluff)_ marks their pink card as the lowest-possible, useful value it could be (given the team's preceding actions).

That's not how normal bluffs work though. After a bluff, the receiver marks their card as all the matching one-away-from-playable cards + all of the N arbitrary bluffs that were previously agreed upon before the game started (3 bluff, critical 4 bluff, etc.)

Zamiell: Trash bluffs exist, and trash bluffs are done with 1's, so correct me if I am wrong, but the "lowest rank" of the card isn't really relevant, its the N-away-from-playable part.

NZadeh: would it be clearer to hyperlink the text to the bluff docs? https://hanabi.github.io/docs/level-11#the-bluff

I used the word "standard" to not have to worry about all of those other bluffs. is there another word you'd like better?

The "lowest rank" matters when determining the possible values of the target of a bubblegum move. For example, on turn 1, Alice gives Cathy a "2" clue focusing slot 1. Bob blind plays red 1 from slot 2. What is Cathy's note?

"Logically" it could be i3, i4, or i5 (not i1 as Bob would have played from slot 1, and not i2 as Bob would have played from slot 1). We agree that it's the lowest rank of these, in this case i3.

docs/variant-specific/pink.mdx
```diff
@@ -292,7 +292,7 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 ### The Bubblegum Bluff (Two-Away-From-Playable Form)
 
 - Normally, _Bubblegum Bluffs_ are always performed by touching _one-way-from-playable_ pink cards.
-- Rarely, it is possible to perform a _Bubblegum Bluff_ on a _two-away-from-playable_ pink card. However, this can only be done in a specific situation where if the clued card to be _one-way-from-playable_, it would not work out.
+- However, it is sometimes possible that the lowest-possible, useful value is actually _two-away-from-playable_.
```
NZadeh, line 295: just a heads up that the language "lowest-possible, useful value" here might be incomplete/lack context if we aren't also taking the rewrite suggestion for the first line.

NZadeh: "...lowest-possible, useful value of a focused pink card is actually..." ?

PR #1062, opened 2023-01-21
Update null 5s chop move to include playable

EhDrone:
Update the null 5s chop move to include playable + phantom playable cards. Based on discussion seemed like there was agreement of those who participated that being able to 5s chop move a playable card or phantom playable card is beneficial with 5s null chop move. Open to additional input why we want to restrict it specifically to criticals.


Zamiell:
ok, seems relatively uncontroversial

PR #1065, opened 2023-02-04
Added 2 Charm Conventions

Jacob-Lisner:
Added the Unnecessary Trash Lie Charm and the Kick Out Charm to Other Special Moves to Pink conventions.


Zamiell:
read https://github.com/hanabi/hanabi.github.io/blob/main/misc/convention-changes.md

PR #1069, opened 2023-02-22
Trash card does not end the early game (lvl 2)

thomas-lehericy:
Added in the definition of trash that discarding trash card does not end the early game, as this does not seem to be present anywhere in the doc. It could be clearer in the "early game section" of lvl 1, but at this point the notion of "trash card" is not introduced.


thomas-lehericy:
I would suggest similar changes concerning
- sarcastic discard
- gentleman discard

Zamiell:
its in level 9: https://hanabi.github.io/docs/level-9/#the-early-game-severity-1-stalling

thomas-lehericy:
Missed it, my bad!

Zamiell:
i linked it now from level 1 so that it is easier to find

PR #1071, opened 2023-02-26
Update muddy-rainbow-cocoa-rainbow.mdx

thomas-lehericy:
The original formulation is misleading: it mentions the "slot number", whereas what is meant is the position of the card. This position is determined, from brown tempo clue conventions, from **right to left**. On the other hand, the original formulation leads to think it should be determined from **left to right**, as when counting slots.

Here is my attempt to resolve this possible misinterpretation.


Zamiell:
very good pr thank you!

PR #1074, opened 2023-05-02
Adding 5 number ejection to sections where it is missed.

Stuurrm:
As a side note the example in docs/examples/5-number-discharge.mdx no longer describes a discharge but an ejection. I do not happen to have an actual example of a 5 number discharge on hand.


Stuurrm:
It has been fixed

Zamiell:
> As a side note the example in docs/examples/5-number-discharge.mdx no longer describes a discharge but an ejection. 

can you change the example so that it properly describes the correct thing?
otherwise delete the example

Stuurrm:
As said I do not actually have an example to give so deleting it.

Zamiell:
Excellent, thank you

PR #1078, opened 2023-05-20
Update convention-goals.md

vEnhance:
"to only to maximize" -> "only to maximize"

I also changed "modular mathematics" to "modular arithmetic" since I'm here anyway :P


Zamiell:
thanks

PR #1082, opened 2023-05-30
Parenthetical note on lv2 challenge q3

vEnhance:
This is possibly too much info. Feel free to close if so.

On the other hand, I actually did get this question from a level 2 player on her first day of reading the document. (Hence my question in the Discord yesterday night.)


Zamiell:
i think it is too much info

PR #1083, opened 2023-05-30
Fixes #1037 (shout + tocm)

### LINKED ISSUE #1037: Shout tocm should mean something special
Fairly often, multiple trashes will be clued as part of a Trash Bluff, and the recipient will already have a playable.

This means they can shout and tocm at the same time. Say slots 1 and 2 are known trash, and Alice discards slot 2 while having a known playable. There are currently two sensible interpretations:

1. Bob double-cms
2. Cathy cms

Since a shout protects all of Bob's cards for a turn, I personally believe that 2 has more utility than 1 --- the cm'ed player should skip over. Moreover, Bob is NOT put into a scream discard situation and may discard.

Either way, we should pick one of these interpretations and run with it.

vEnhance:
Shout + TOCM should chop-move cathy


Zamiell:
it hasn't been strongly defined before now but sure

Zamiell:
did you mean slot 3?

Zamiell:
i updated the text to be more clear

PR #1091, opened 2023-07-18
Update omni.mdx

vEnhance:
crosslink pink conventions


Zamiell:
thx.
do you also want to investigate why ci is failing?

vEnhance:
Looks like https://stackoverflow.com/questions/76708329/docker-compose-no-longer-building-image-attributeerror-cython-sources

upgrading pyyaml should fix it, let me try

Zamiell:
yeah might as well upgrade all the deps to latest while you are at it

Zamiell:
thanks!

PR #1092, opened 2023-07-18
Bump PyYAML to 6.0.1

vEnhance:
(just to check if the action succeeds after this)


Zamiell:
looks like it worked, yay

vEnhance:
Cool, I also bumped svgwrite/black while I was here, if this passes CI we can probably just merge it.

Zamiell:
thanks evan!

PR #1093, opened 2023-07-19
Simplifies the process for determining the focus of a clue

Egroegw:
We can replace the four-step process for determining the focus of a clue with a shorter yet logically equivalent two-step process:

> 
> 1. If the card on chop is touched, then **the focus is on the chop**.
> 1. Otherwise, **the focus is on the left-most new card**. 
>     - If no cards are new, then **the focus is on the left-most re-touched card**.


Zamiell:
those look like 3 steps to me

Zamiell:
with less verbosity we lose clarity, it isn't clear to me that the tradeoff is worth it

PR #1098, opened 2023-08-09
Update pink.mdx

newduke:
Clarify pink trash moves

### DIFF
--- a/docs/variant-specific/pink.mdx
+++ b/docs/variant-specific/pink.mdx
@@ -387,8 +387,8 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 
 ### Trash Discharges and Charms in Pink
 
-- In Pink, (rank) nonfocused cards will have soft good touch. What happens after we clue a trash pink where a single focus would imply bluff (and there are outstanding unknown good pinks)?
-- If **any** of the nonfocused cards are good, this will become a _Unknown Trash Discharge_.
+- In Pink, (rank) unfocused touched cards will have soft good touch. What happens after we clue a trash pink where a single focus would imply bluff (and there are outstanding unknown good pinks)?
+- If **any** of the unfocused touched cards are good, this will become a _Unknown Trash Discharge_.
 - Otherwise (all touched cards are trash), this is an _Unknown Trash Charm_.
 
 ### Pink Precedence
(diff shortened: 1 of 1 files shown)


Zamiell:
Can you make it pass ci

Zamiell:
still doesn't pass ci

newduke:
It passes CI

Zamiell:
i dont understand the clarification. is this a brand new convention?

newduke:
It changes nothing. But it was confusing before as to what happens in these situations. Notably, players were confused about bluffs in pink because they can be confused with utc.

Zamiell:
it looks like it changes the trash bluff rules.
what you clarified isn't how we play.
trash bluffs !== UTD

Zamiell:
have you read this section? https://hanabi.github.io/docs/variant-specific/pink/#trash-finesses--trash-bluffs-in-pink

newduke:
Hrm fair. But what if all trash? Should still be utc no?

Zamiell:
yes, it would be UTC, but does that need a clarification? it just works the same as No Variant, i dont think there needs to be a separate section to clarify just that (?)

Zamiell:
can you update the pr?

Zamiell:
thanks

NZadeh:
for what it's worth, this is definitely not how most would people play the clue (though people's interpretations may have conflicted with some of the previous language).

for example, a 1 clue touching i1, r1, b1 that is intending to get the final 1 to play from finesse.... is now supposed to be a charm? I don't think that's how anyone will play.

The perception is likely to be that the doc just looks _more_ wrong now, not less.

I suspect a happy compromise would be something like: a player is promised all trash, unless "fixed" to be shown prink cards. But I haven't thought through the exact convention.

NZadeh:
It's also worth mentioning that this is lvl 21 in pink, but a charm isn't introduced until level 24...

Zamiell:
> It's also worth mentioning that this is lvl 21 in pink, but a charm isn't introduced until level 24

can you submit a pr to add the standard boilerplate "if you are level 24, then" part?

Zamiell:
> for what it's worth, this is definitely not how most would people play the clue

then someone should start a github thread to change the conventions.
i'm not just going to arbitrarily change the doc because some of the new players are going rogue and playing differently

NZadeh:
haha I think it's more that the old players are going rogue. new players have little reason to.

NZadeh:
> > It's also worth mentioning that this is lvl 21 in pink, but a charm isn't introduced until level 24
> 
> can you submit a pr to add the standard boilerplate "if you are level 24, then" part?

what does the move mean before level 24? bad touch trash bluff/finesse?

Issue #959, opened 2022-08-01
Summary of the conventions


Romain672:
Hi,
How about creating a page 'summary' which would take most conventions, and wrote one line/~10 words to explain it.
It would be for players who know at 80/90% the conventions, and just need a quick refresh to know what to answer/what a specific clue would do?

Romain672:
Like, if you open the page, and search for '5-pull', you will get some quick summary to what happen depending of which card clued.
Actually, if you search for '5-pull' on the doc, you will get like 50 lines of 5 differents conventions. Then need to scroll down to check what you need exactly.

Zamiell:
ok, you can submit a pr

Romain672:
Okay I will do it.
If anyone want to help I started it here: https://docs.google.com/spreadsheets/d/1LECCSBMPYOTKugRKxk69iqRG7fF43-tIJkU4cVSufMQ/edit#gid=0

It will became a page at the end.
Feel free to do any modification for clarity, I just wrote the first sentence I got, I didn't double checked anything, I will do it later.

Romain672:
Since I got no answer, here is where I'm at, without the excel document:
https://textup.fr/643899Me
I really would want some early feedbacks though.

Zamiell:
seems fine

Romain672:
Just add my text into a new page. There still need to do everything else (like doing the links between the page per example), if anyone else can do it: #990

Zamiell:
i'm not a big fan of this since whenever we shuffle the levels around, or add new conventions, or change anything, we have to also change the summary

Zamiell:
which is more maintenance work for me

Romain672:
I guess there should be javascript or similar to get the Titles of sections added and removed automatiquely?
But that's a lot of extra work.

Zamiell:
ok i will close this for now

PR #1099, opened 2023-08-10
Summary of the conventions

vEnhance:
WIP, this is like half done, but throwing out a draft now to get impressions from people.

In short, this is supposed to be a high-level overview of the commonly used conventions in basic levels 1-25.

Similar to #959/#990 but organized conceptually rather than directly by level (and we have a script to shuffle levels now), and only covers a subset of the basic conventions (and not the extras). So I think it should not be too much maintenance work (and I can certainly help with that).


vEnhance:
OK done writing the cheat sheet!

Zamiell:
can you rebase on main?

vEnhance:
> can you rebase on main?
done

(also, in `learning-path.mdx`, the big diff is just because i added the word "level" in front of the number, none of the lists got edited)

Zamiell:
yeah thats why i updated main

Zamiell:
i can see the diff more clearly now

docs/learning-path.mdx
```diff
@@ -24,6 +24,7 @@ For information on how variants tie in to the levels, see the [variant summary](
 ### Level Summary
 
 The following is a quick summary of the levels.
+You can also see a [cheat sheet of many of the main conventions](summary.mdx).
```
Zamiell, line 27: this won't appear on a newline, is that intended? if not, then put it on the same line

PR #1100, opened 2023-08-10
5CE and 4charm only work on new 4's/5's

vEnhance:
e.g. if a card was previously clued by 5 and then I fill it in with color, that does not trigger a 5CE

I _think_ this is how we play already. Correct me if I'm wrong. :P


Zamiell:
this is correct

PR #1107, opened 2023-08-16
Update variants.md path

vEnhance:
Because of https://github.com/Hanabi-Live/hanabi-live/issues/2872


Zamiell:
thanks.
(i realized that only specific markdown files should be in uppercase, not all markdown files)

PR #1111, opened 2023-09-08
Update level-10.mdx - clarify Sarcastic Finesse

J-Fricke:
Inform that the card used must be unknown it's in the clue giver's hand


Zamiell:
thanks!

PR #1115, opened 2023-09-30
Update level-3 docs

therealtod:
With this update we introduce an example of fix clue which doesn't involve touching the card which we intend to fix. This kind of scenario is currently not conteplated by the examples in the docs for level 3.

ID of the original thread on the H-group discord server: 1157308972542464081


Zamiell:
thanks

PR #1116, opened 2023-10-01
Add udd to level 16

unsoftcapped3:


Zamiell:
its intentionally not in level 16.
what is the point of this pr?

unsoftcapped3:
many people have requested it to be part of level 16. Almost all level 16 players know it. This is to formalize udd being turned on in level 16 games.

unsoftcapped3:
If you make a poll I think most people would agree this should be in level 16

Zamiell:
if you want to do a poll, the correct way is to open a github issue, can you open an issue?

PR #1120, opened 2023-10-31
Clarify Elimination Play Clue with multiple elim cards touched

bnordli:
As discussed here: https://discord.com/channels/140016142600241152/1164236186475298828/1164236934999195778

(We encountered the exact same question while playing today.)

For brevity, I changed the current example. The case where only one card with elim notes is touched is covered at the top of the chapter.


Zamiell:
thanks

PR #1124, opened 2023-11-11
For White, include number 5 clues as Always Loaded

Egroegw:
We can state that the Always Loaded Principle applies to the White 5 as well. The Gray convention page states "normally, the Always Loaded Principle applies to white cards" and that the ALP doesn't apply to 2s and 5s, implying that the ALP does apply to 5s in White.

### DIFF
--- a/docs/variant-specific/white.mdx
+++ b/docs/variant-specific/white.mdx
@@ -9,7 +9,7 @@ These conventions apply to any variant with a white suit (i.e. touched by no col
 
 ### The Loaded Play Clue Assumption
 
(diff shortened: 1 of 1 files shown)


Zamiell:
thanks

unsoftcapped3:
this is literally the dumbest thing ever

unsoftcapped3:
in my hundreds of games of hanabi I have not once seen a single player user loaded 5s, ever.

newduke:
I've played with w5 lpc after low score phase, but certainly never "always loaded."

Issue #730, opened 2021-08-03
Convention Proposal: The Always Loaded Principle


Lel0uch-H:
There has been a lot of confusion regarding White Loaded play clues. Similarly, in gray/dark prism/etc. variants we say that "dark prism 3 and 4 saves are loaded" but do not specify the behavior about what to do when a non-dark prism 3 or 4 is early saved.

I believe the goal of such conventions is just to have tempo on cards where it would be impossible to, otherwise. But having such discrete rules for such variants fails the goal of simplicity and uniformity.

I suggest we cut the jargon and just define it as "all rank saves are to be treated as loaded play clues in a variant where any rank clue to chop _usually_ looks like a save" (and maybe even the counterpart of this rule for color clues for variants like Brown-Fives let's say, but that is a topic for another PR). This just looks clean from an aesthetic standpoint and would not cause any disagreement in game about clue interpretation.

dobi815:
This is how we have been playing I think. What would it take to firm this up? Should just make a PR?

1. I don't see any downsides to treating a 3 or 4 on chop for a critical 3 or 4 additionally as a play clue regardless of whether its from the special suit or regular suit. So definitely yes.

2. The absence of this convention results in special casing for when the clue receiver could potentially interpret the clue as a play (and thus contextual use of that in higher levels). Its more confusing than less.

Zamiell:
> But having such discrete rules for such variants fails the goal of simplicity and uniformity.

could not agree more. a well written issue, and you have my support

Zamiell:
> This is how we have been playing I think. What would it take to firm this up? Should just make a PR?

read [convention-changes.md](https://github.com/hanabi/hanabi.github.io/blob/main/misc/convention-changes.md) to see what the process is

Lel0uch-H:
I have expanded the scope of the issue to talk about both such rank and color clues.

Zamiell:
@Lel0uch-H shouldn't your PR have also deleted some other sections in the doc?

PR #1126, opened 2023-11-14
No loaded 5s ever.

unsoftcapped3:


Zamiell:
please read https://github.com/hanabi/hanabi.github.io/blob/main/misc/convention-changes.md, which goes over how we change conventions

unsoftcapped3:
You literally just merged his change without getting feedback from anyone else.

unsoftcapped3:
did a single person approve it?

Zamiell:
afaik this was just a clarification of an existing convention, not a new convention. from what im reading of ALP, it should apply equally to white 2 and white 5, correct me if i am wrong

unsoftcapped3:
It does not apply to white 5. Have you ever seen a game where anyone played with loaded 5s? I haven't. I play with a lot of people in H group and no one plays with loaded 5s when doing white.

unsoftcapped3:
this is not "just a clarification of an existing convention". It introduces a new convention (all 5s are loaded) that we were not playing with before. This is a new convention and was added without any consensus.

Zamiell:
right now i'm going off what the doc says, can we restrict the convention to that please?

yagamiblack:
The doc for White explicitly mentioned rank clues with 2, 3, and 4 before the aforementioned pull request, therefore adding white 5 *IS* a convention change, and one that was not approved without any feedback.

Zamiell:
afaik that line is just a reiteration of what ALP means, right?

unsoftcapped3:
The line "Since it can be hard to get white cards on chop to play, the Always Loaded Principle applies to save clues given with number 2, number 3, and number 4" is not a reiteration of what ALP means. It is a specific added convention to the white variant.

Egroegw:
I suggest then that we clarify the ALP and, if necessary, change the Gray convention page to stop implying that the ALP applies to 5s.

unsoftcapped3:
sure I can do that

yagamiblack:
Also do note that as the majority of H Group players do not play with loaded number 5 clues in White, the previous pull request will only cause confusion when newer players reading White conventions play with more experienced players, hence the need for more community discussion.

jack67889:
Having the convention apply to 5s is at least not in the spirit of the reason to introduce the convention [in the first place](https://github.com/hanabi/hanabi.github.io/issues/730). But I agree it's very unclear and it has been raised multiple times by several people since the introduction, but I've never seen any clarification. I've never understood what the intention of the convention is.

Problem is also that the example suggests it only applies to cards that could be currently playable white cards (like the convention that preceded it), but that is also left unclear. In a [discussion on Discord](https://discord.com/channels/140016142600241152/456214043351580674/928537046849576960) it was suggested that the intention is indeed to only differentiate between play/save, like the example, which makes the wording of the actual convention way too broad imo. If that's the intention you're probably better off to restrict it to only possible playable white cards (with or without self finesses?). A stricter interpretation would also leave open the possibility for rank self-finesses instead of (any) saves. While that convention was deleted right before ALP was introduced, so that also seems unintended to me.

Zamiell:
thanks jack for the summary, i'll open a new issue

PR #1129, opened 2023-11-19
Match prompt-question-1 text to image

demipixel:
The image shows both the 4 and 5 already prompted, but the text only says 5 is. Fixed the text to match the image (and solution)

https://hanabi.github.io/docs/beginner/prompt-question-1


Zamiell:
thanks

PR #1133, opened 2023-11-26
Update level-25.mdx

vEnhance:
Removes the "info asymmetry" line now that we have always unnecessary


Stuurrm:
That line still applies as UTD can still be considered necessary.

vEnhance:
oh, good point

is UTD the only one?

Stuurrm:
UTD, UTC and prophetic discharge I believe are all the ones.

PR #1137, opened 2023-11-27
Update special-fives.mdx

David707x:


Zamiell:
you spelled color wrong and forgot the italization, but otherwise thanks

PR #1140, opened 2023-11-30
Update level-8.mdx

fpvandoorn:
With 3 clues, cluing cannot cause a loss. If Alice doesn't have 4s, then Cathy will draw one when playing r3 and the team still wins.


Zamiell:
thx

Issue #1141, opened 2023-12-01
Consolidate special-ones/-fives pages?


Egroegw:
Would it be good to consolidate all the conventions for special-ones/-fives onto two pages: (1) special-ones and (2) special-fives? This would make searching easier.

We currently have pages on
- Omni-ones
- Pink-ones
- Rainbow-ones & Rainbow-fives
- Deceptive-fives
- Null-fives
- **Special-fives**

On the **Special-fives** page, we have more sections on
- Rainbow-Fives & White-Fives
- Pink-Fives & Omni-Fives & Light-Pink Fives
- Brown-Fives
- Muddy-Rainbow-Fives

Zamiell:
sounds fine to me, do you want to submit a pr

Zamiell:
in next next PR remember to put "Closes #1141" in the description, then this issue would be automatically closed

PR #1142, opened 2023-12-02
Consolidate special ones and fives, add mud clues to omni, rainbow fives, muddy-rainbow fives and omni-fives

Egroegw:
This PR fulfils #1141 and also adds mud clues to omni, rainbow fives, muddy-rainbow fives and omni-fives. (Please ignore the first two commits, which I've manually reverted.)


Zamiell:
the number of commits in a PR don't matter here, since we always squash before merging

Zamiell:
your PR doesn't pass CI, did you test your changes?

PR #1143, opened 2023-12-02
consolidate special-fives and -ones

Egroegw:
Addresses #1141 and adds mud clues to mud clues to omni, rainbow fives, muddy-rainbow fives and omni-fives.


Zamiell:
nice thanks

PR #1149, opened 2023-12-08
UDD out-of-level fix

Egroegw:
Since UDD was moved from max level to level 16, some parts of the UDD section are out-of-level. I've made minor changes to fix this.


Stuurrm:
Without question even before lvl 18 players should mark notes that they hold the other copy. The elim notes actually do nothing even at max other than just remind one of those cards is the discarded card.

kesslermaximilian:
The only thing I could see why we want to rephrase this is because the term 'Elimination Notes' might not be clear. But 'marking they have the other copy' is very clear also at level 16 of course.

Zamiell:
i added some edits, thanks

Issue #1032, opened 2022-11-09
Convention Simplification: Trash Bluff Unnecessary Assumption


pianoblook:
Zam and I briefly discussed this a few weeks ago, but I forgot about it until earlier today - woops! 

Currently it's often impossible for Bob to know whether or not a _Trash Bluff_ will end up CMing the clue receiver. This desync can lead to headaches - and even bombs. This remains true, and can even get more chaotic, amongst experts, when it becomes a 'leveling game' where everyone needs to start scrolling back through the replay to determine whether or not Bob **should** be able to piece together the truth one way or the other... It's a mess.

So anyway, below is the _Unnecessary Assumption_ Principle, taken from this [doc](https://docs.google.com/document/d/1dLGcc1iZ_NrxC2rcHDlM6pH_Qna_FrIF8zXMJk2nZGw/edit#heading=h.6t144y14pc2m). We've been playing with this for over a year now, and it's been a HUGE help. Simple and powerful.

----------

## Trash Bluff Unnecessary Assumption 

- When a _Trash Bluff_ is performed, we agree that Bob should always assume that Cathy will respond as if it was an [Unnecessary Move](https://hanabi.github.io/docs/level-25/).
- Therefore, if Bob sees that this _Unnecessary_ interpretation would lead to a clearly undesirable outcome, he can know that he’s actually being called to [Signal Shift](https://hanabi.github.io/docs/extras/signal-shifting/) appropriately.
> - If the Unnecessary component would lead to Bad CM, Bob must Eject.
> - If it would lead to an impossible _Trash Push_, Bob must either _Eject_ (as a _TPE_) or _Discharge_ (as a _TPD_).
- Notice how Bob’s Ejection or Discharge **still serves the same purpose as a standard _Finesse_**: the fact that he responds with **any** blind-play immediately identifies the clued card(s) as trash.

NOTE: If multiple 1s are clued, [Junk Discharge](https://hanabi.github.io/docs/extras/discharges/#the-junk-discharge-for-1s) takes precedence unless the unfocused 1 is clearly stale. In Pink variants, [Turnabout Ejection](https://hanabi.github.io/docs/variant-specific/pink#the-turnabout-ejection-for-1s-and-5s) also takes precedence.

### Implications for Cathy
- Since Bob’s response to all Trash Bluffs are now already determined by the nature of the _Unnecessary_ component, it doesn’t actually matter whether or not it actually **was** _Necessary_.
[...]

PR #1153, opened 2023-12-18
Fixes for Unnecessary Trash Finesse/Bluff

argothiel:
There was some obsolete wording left after the #1032 implementation. This commit cleans it up.


Zamiell:
thank you!

PR #1154, opened 2023-12-19
Update level-23.mdx

newduke:
update TOCM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - run: npm run build # It is faster to rebuild than to download/upload the build output.
       - uses: actions/upload-pages-artifact@v2
         with:
-          path: "./build"
+          path: ./build
       - uses: actions/deploy-pages@v2
 
   auto-format:
--- a/docs/level-16.mdx
+++ b/docs/level-16.mdx
@@ -149,7 +149,7 @@ import UnknownDupeDischarge2 from "@site/image-generator/yml/level-16/unknown-du
 - _Unknown Dupe Discharges_ only apply if **the two duplicated cards are in the same person's hand**.
   - It is explicitly illegal to perform an _Unknown Dupe Discharge_ that duplicates a card in someone else's hand. If this happens, the clue must have some other meaning.
 - Remember that after an _Unknown Dupe Discharge_, the player who received the clue is supposed to _discard the focus of the clue_.
-  - However, in the case where only two new cards are touched, then the player who received the clue knows that both of these cards must be the same. In this special case, they can discard the non-focused card to cause a _Trash Order Chop Move_.
+  - If the player discards another card other than the focus of the clue, it is to be treated as a known mistake.
 - In order for players to determine whether or not a _Double Finesse_ or a _Triple Finesse_ is happening, they should use the same _two-or-more-blind-plays_ rule that applies to _5 Color Ejections_.
 - For example, in a 3-player game (similar to the previous example):
   - It is the first turn of the game and nothing is played on the stacks.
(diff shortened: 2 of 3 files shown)


newduke:
I propose adding the line for the *exact* situation where alice has 2 known playable global dupes. Doesn't it follow from the guidelines of TOCM that alice can cm Cathy here?

kesslermaximilian:
Well, it is the case if these are globally known duplicates, but as I outlined above, that is something different than being UDD duplicates, because the status if they are global is complicated.

newduke:
If the cards are play clued (prompted) then I think it is pretty clear they are globally known

newduke:
Actually, scratch that. The main point is if Alice discards the wrong dupe, this proves they were globally known. This is true for all UDD TOCM situations.

Zamiell:
> This is true for all UDD TOCM situations.

@kesslermaximilian is that true?

kesslermaximilian:
Yes, that's the point I was outlining above, if you relplce "globally known" by "privately known".
And that's the complication why just adding duke's comment is a bad idea.

The only satisfactory solutions I know of are a) ignore this or b) the doc I linked, which I don't think is useful for regular H becasue it's very long/complicated.

That being said, even without this doc, the linked convention does seem to be best practice considered among several max-lvl players I taked to before while writing up that convention.

Zamiell:
I updated the PR to say this:

> - Note that it is explicitly illegal to perform a _Shout Discard Order Chop Move_ if the player has exactly two trash cards that were touched by an _Unknown Dupe Discharge_. In the case, if a player discards the cards out of order, it is to be treated as a known mistake.

kesslermaximilian:
No, that it also not how it works. The whole point is that if you have cards from a UDD, you *can* tocm with them by discarding non-focus first. The tricky thing is that this might differ from the usual left-to-right rule.

It is legal, and people use it all the time.

Additionall, the interaction of having UDD dupes *and* further known trash in hand is tricky and not regulated in general H.

kesslermaximilian:
what is this force-push duel here?

Zamiell:
sorry for the spam, im fixing a bug in CI related to PRs, so im using this PR as a test bed to see if it works, but anyways it should be fixed now

kesslermaximilian:
I see

Zamiell:
what about this:

> Note that it is explicitly illegal to perform a Shout Discard Order Chop Move using a card that was touched by an Unknown Dupe Discharge. While it is normally legal to perform a _Trash Order Chop Move_ from a UDD that touches exactly two cards, you cannot use it to shout.

kesslermaximilian:
I dont see why we want this in the first place. Discarding non-focus demonstrates the UDD is known, so we can totally TOCM + Shout with UDD dupes. You just have to be aware that - as with all UDDs - it is focus first not leftmost first.

newduke:
I am still in favor of the original wording.  I don't see how there is any confusion if Alice has a known play of the left dupe. If it is a dupe, and playable, both cards are known.

newduke:
Yes this seems like a mistake.  My wording wasn't intended to change anything, just to clarify how we already treat TOCM with UDD.  I believe @kesslermaximilian and I agree that my original wording, perhaps amended to add "the dupe which was not the original focus of the UDD" is correct.

Zamiell:
(for context, the original PR was "- Note: this also applies to known dupes from an _Unknown Dupe Discharge_.  Both cards are considered trash for the purpose of _Order Chop Moves_.")

Zamiell:
@kesslermaximilian can you construct an alice and bob story for newduke to illustrate why adding this bullet point to the docs is a problem

ricardodd2:
I just think Newduke wording of the pull request was unfortunate. I know he intended that, similarly to trash order chop move, you Shout with the focus of the UDD and you Shout Discard Order Chop Move with the non focus card.
When there is a single UDD+globally known trash it seems logical to me that the order of discard should be globally known trash or focus of the UDD left to right and the none focus of UDD last. When multiple UDD occur, I propose that the focus of UDD's/known trash is left to right, while the none focus of UDD have same precedence. This way there is no risk of desincronization.
For example, In a 3 player game, if Alice has g2g2r3r3b1 and received a blue clue, a red clue and a green clue focusing slot's 5, 4 and 2. Discarding slot 2 would cm Bob, Discarding slot 4 would cm Cathy and discarding either slot 1 or slot 3 would cm Bob twice.

Zamiell:
closing in favor of #1172

PR #1157, opened 2023-12-25
I wrote something lol

vEnhance:


Zamiell:
i fleshed this out a bit and changed the overall structure somewhat, do you want to review my edits?

Zamiell:
agree

Zamiell:
updated now

EhDrone:
> > i fleshed this out a bit and changed the overall structure somewhat, do you want to review my edits?
> 
> I think looks OK to me, and a lot more developed. (The story about BGA players moving to Keldon Jones website is actually new to me, I didn't know that part of the history.)
> 
> I personally think it might not be good to call out smallman in particular. I think a lot of people who read this essay might be people who are not members of the H-Group (e.g. prospective members) and I think it could be a turn-off for people if their first impression of the H-Group was that we don't tolerate a certain playstyle from a specific player.

Was going to say something similar. I'll take a pass at the latest edit today and maybe make some updates or add something myself

PR #1162, opened 2024-01-09
Update pink/light-pink/omni fives

gw12346:
Motivated by debate in a post-game analysis (https://hanab.live/shared-replay/1092514#1) to clarify play clues on fives.


Zamiell:
see my edits.

accepting this pr depends on the resolution of #1164

gw12346:
Full support for your edits, especially for removing the boilerplate "this section applies only to ... fives".

Zamiell:
> Do we need to split the pink/light-pink/omni fives section into two then?

@gw12346 do you want to take care of that in a follow-up PR?

PR #1166, opened 2024-01-14
Added python command requirement

Aznkad:
I added a check that the `python` command does exist in the `install.sh` script, since one can only have the `python3` command installed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,8 +29,13 @@ echo "Complete."
 
 echo "Installing Python dependencies..."
 cd "$DIR/image-generator"
-if command -v "pip3" > /dev/null; then
-  pip3 install -r requirements.txt
+
+# Check that `pip3` and `python` commands are present
+if (command -v "pip3" > /dev/null; command -v "python" > /dev/null) then
+   pip3 install -r requirements.txt
+elif command -v "python3" > /dev/null; then
+  echo "Error: Your Python program appears to be assocaited with the \"python3\" command. The \"python\" command is required for this website."
+  exit 1
 elif command -v "pip" > /dev/null; then
   # Check to make sure that pip is for Python 3, not Python 2
   if [[ ! $(pip3 --version | sed -n 's/^.*(python \(.\).*/\1/p' | grep 3) ]]; then


Zamiell:
i dont think this check does anything because pip3 implies python3

Aznkad:
The first condition can fail either if the python command does not exist or the pip3 command. in the first case I think this is helpful to detect whether, though python is not here, python3 is here or not

Zamiell:
it looks like you are using a semicolon as an `and` operator, is that intentional?

Zamiell:
also this fails ci, make sure your prs pass ci before submitting

Aznkad:
I'm sorry i'm not used to CI, how can I check if i pass them ?

Zamiell:
run lint.mts

Aznkad:
looks like the last commit passes ci

Zamiell:
i think the current PR is bugged because the script would fail if they had `python3` + `pip3`, right?

Aznkad:
Yes it'd fail because if they only have python3 they should install python. the other solution is to replace every occurrence of `python` in the code with a check of which python's command is installed

Zamiell:
> the other solution is to replace every occurrence of python in the code with a check of which python's command is installed

it is already like that right?

Aznkad:
The last time i checked it was not the case, for instance line 46 in `/image-generator/index.js`

Zamiell:
should be fixed in this commit: https://github.com/hanabi/hanabi.github.io/commit/e31ee4f41cc2d4632714d72839ed60e79b108295

anywhere else?

Aznkad:
should be sufficient, i will try on my machine

Aznkad:
did you push ?

Zamiell:
push?

Aznkad:
it might be a git-skill issue but i cannot git pull the change to test on my machine

Zamiell:
why not

Aznkad:
"already up to date"

Zamiell:
probably because you are pulling from your branch instead of the main repository

Zamiell:
when you fork the repo, it stays forever at that point in time, it is your responsibility for keeping your fork up to date with the upstream

Aznkad:
seems to be working! i'll close the pr

PR #1167, opened 2024-01-15
Added self-bluffs

Aznkad:
By playing with some people I found out *Self-Bluffs* were a thing : they are mentioned for instance in `extras/special-bluffs.mdx`. Since _Self-Finesses_ have their own dedicated section in the level 2, i guessed _Self-Bluffs_ deserved their own too, hence I wrote one. 
I did not invent an example, I took the one from `extras/special-bluffs.mdx`.


Zamiell:
self color bluff is different than self bluff, and doesn't belong at level 11. i will edit your pr

Zamiell:
thanks

PR #1168, opened 2024-01-17
Clarifications in Level 16

Aznkad:
List of changes:
- Proposal of clarifications on the example with an _Out-Of-Order Finesse_, since this only appears at level 21 and can be confusing for new players
- Changed a _6_ to a _5_ after the list of authorized duplications
- Added the mention of the _Layered-Finesse_ to the list of finesses allowing duplications. This is because the example originally said that Alice expected a _double-finesse_, however given that the clue is a color clue and there are only 3 players this would be nonsensical. I also modified that.


Zamiell:
> Added the mention of the Layered-Finesse to the list of finesses allowing duplications. This is because the example originally said that Alice expected a double-finesse, however given that the clue is a color clue and there are only 3 players this would be nonsensical. I also modified that.

it is a double finesse though, a layered finesse is something else

Aznkad:
my bad, i thought that as soon as the cards were stacked in the same person's hand it was a layered

Zamiell:
no, layered means something of a different suit

PR #1169, opened 2024-01-22
Update dark-omni-gray-pink.mdx

ricardodd2:
Since dark omni has only one card of each rank, soft good touch principle loses its purpose

### DIFF
--- a/docs/variant-specific/dark-omni-gray-pink.mdx
+++ b/docs/variant-specific/dark-omni-gray-pink.mdx
@@ -9,7 +9,8 @@ These conventions apply to any variant with a dark omni suit or a gray pink suit
 
 ### Inheritance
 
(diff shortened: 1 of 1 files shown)


Zamiell:
isn't this self evident?
i dont think this makes the docs better, it just kind of adds extra noise that makes this line harder to read

ricardodd2:
They asked me to do the pull after game: https://hanab.live/replay/1099073#35. Where one person was claiming that b4 shouldn't play on turn 35 because of soft good touch principle. I think that not writing anything makes it unclear. If adding it the way I did makes the text hard to read I suggest trying to find another way to communicate it. What about adding an extra line like "- Exeption: Soft Good Touch Principle no longer applies to dark omni".

Zamiell:
ok

PR #1171, opened 2024-01-30
Add known trash discharges

Aznkad:
This convention proposal originated from this replay : https://hanab.live/replay/1103041#55. We were arguing on whether or not this move was an UTD or not, and @tulioz on the discord told us that it indeed was an UTD.
However I find this very unclear since the reasoning Kernel has to do in order to blind play his slot 3 is clearly different from the one needed during a classical UTD.
Thus I find it best to present it as a new move called "known trash discharge". Feel free to present things in another way, however if @tulioz is right and the move in the replay _is_ a discharge, then I think it really should be explained somewhere in the doc.


Stuurrm (changes requested):
It could be more compact if instead of adding a new convention a meantion of hits would be added to the "The Unknown Trash Discharge (2-for-1 Form) (UTD)" section. These kind of clues are very common but I guess there isn't an explicit mention of it.

Zamiell:
i agree with stuurm, please make that change

Aznkad:
I have some questions:
- What do you mean by "mention of it in the UTD (2-for-1)" ? should I remove the example ? as well as the mention of signal shifting ?
- Why is it a UTD n-for-1 ? In the example I give, if the red 1 does not exist this is not a UTD, and if i understand correctly a UTD n-for-1 is a UTD which touches more than 1 card

Aznkad:
Sorry to ping again, but did you see my comment?

Zamiell:
> should I remove the example ?
you can put the example in the 2-for-1 seciton

>  as well as the mention of signal shifting ?
yes remove that

> Why is it a UTD n-for-1 ?
because its not different enough to warrant having a different name

Aznkad:
Should be changed!

Zamiell:
you can review my edits

Aznkad:
I agree with your changes, thanks for merging

PR #1172, opened 2024-01-31
Clarify discard order for UDD and consequence for shout discard order

kesslermaximilian:
This replaces #1154. See the discussion on that PR for the reasoning behind this.


Zamiell:
thanks

PR #1180, opened 2024-02-22
Double ignitions

gw12346:
The current wording of double ignitions suggests that Alice can perform a DI that gets Cathy's and Donald's finesse positions even when Bob has nothing to play.


Zamiell:
see my edits

PR #1189, opened 2024-03-24
Add white-ones and self color bluffs

gw12346:


Stuurrm:
I do not know what the concensus on this is. This is a new convention for which I at least have been personally against.

Zamiell:
open an issue for new conventions rather than a PR

gw12346:
No problem! Kappatosity told me that this was standard for white-ones, so I thought there was already consensus.

PR #1194, opened 2024-04-16
Update level 13 (crit color bluffs)

gw12346:
Since critical color bluffs originated from the black 4 bluff, it's always been H-group history that critical color bluffs can also apply to dark 4s. However, it seems like many newer players do not realize this, since the docs don't state this explicitly.

By changing the example from a bluff on red 4 to black 4, we can make this clearer without changing any wording.


Zamiell:
we want all the examples to be for no variant though

Zamiell:
can you just add a bullet point instead maybe?

gw12346:
How's this wording?

Zamiell:
put the example back to no variant though

gw12346:
But the example _is_ back to novar?

Zamiell:
no, it doesn't look like it:

image

gw12346:
Opps! Guess that's what I get for not using the Find function.

PR #1196, opened 2024-04-19
Fixing text to match picture

joshkopen:
Screenshot 2024-04-19 at 4 36 20 PM

Fixing the text in the description to match the image right below it.

It does not substantively change the example as the focus is still the same, but it is inconsistent with what is directly below it.

Alternatively, could change the image, but this seemed easier.


Zamiell:
thx

PR #1198, opened 2024-04-23
Update level-20.mdx

EhDrone:
Adding clarity on Priority 6 - Unknown card. If this is outdated, we need to update the flowchart overall.

### DIFF
--- a/docs/level-20.mdx
+++ b/docs/level-20.mdx
@@ -41,16 +41,6 @@ import TrustFinesse from "@site/image-generator/yml/level-20/trust-finesse.yml";
   - Alice plays blue 2.
   - Bob comes next. Bob knows that normally Alice had a choice and therefore were supposed to play the lowest rank card - red 1 (unless it is a blind-play, or it leads into someone's hand, or it is a 5). Alice **did not** play the card with _Priority_.
   - Bob doesn't see blue 3 in Cathy's hand. This must have been a _Priority Prompt_. Bob plays the leftmost clued 3, and it's blue 3.
-- For Priority 6 - Unknown Card - It is important to understand that playing an Unknown Card does not trigger priority inherently by being unknown. In order for priority to trigger, one must consider all possible options for the unknown card and all possible options must have priority **over** the known card that was played.
-- For example, in a 3-player game:
-  - Alice has a clued + known playable red 2 in slot 2 and a clued playable unknown blue 2 in slot 3. Red 1, Blue 1, and Yellow 1 are all played on the stacks.
-  - No other player has any 3s in hand.
-  - Alice plays the known red 2.
-  - Bob comes next. Bob knows that the priority chart states Unknown Cards have priority over Left-most card. However, Bob also knows that the Unknown card options must all have priority over the known card for priority to take effect. Bob can see that blue 2 and yellow 2 do not have priority over red 2. Therefore, because all options for the unknown card **do not** have priority over red 2, Bob does nothing and this is not a priority finesse.
-- A second example for Priority 6 that _does_ cause a priority finesse, in a 3-player Rainbow game:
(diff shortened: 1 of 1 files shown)


EhDrone:
For ease of discussion if needed - my understanding is that an unknown card only has priority over a known card if all of the possibilities for the unknown card also have priority over the known card. I put 2 examples in the updated page for this to reference when it would and would not have priority over a known card that i'll reflect here:

- For example, in a 3-player game:
  - Alice has a clued + known playable red 2 in slot 2 and a clued playable unknown blue 2 in slot 3. Red 1, Blue 1, and Yellow 1 are all played on the stacks.
  - No other player has any 3s in hand.
  - Alice plays the known red 2.
  - Bob comes next. Bob knows that the priority chart states Unknown Cards have priority over Left-most card. However, Bob also knows that the Unknown card options must all have priority over the known card for priority to take effect. Bob can see that blue 2 and yellow 2 do not have priority over red 2. Therefore, because all options for the unknown card **do not** have priority over red 2, Bob does nothing and this is not a priority finesse.
- A second example for Priority 6 that *does* cause a priority finesse, in a 3-player Rainbow game:
  - Alice has a clued + known playable red 2 in slot 2 and a clued blue card in slot 3. 
  - Alice plays red 2.
  - Bob comes next. Bob knows that the priority chart states Unknown Cards have priority over Left-most card. Bob considers the options for the blue card in Alice's hand as blue 1 or rainbow 1. Because both of those possibilities would have priority over a red 2, Bob plays his finesse position as a *priority finesse* and successfully played red 3.

benzloeb:
I disagree with the first example you gave. Ultimately priority is meant to force people to play cards in the "correct" order and reward us for when we don't. I think it is fundamentally better for Alice to play the unknown 2 because it resynchronizes information of what was clued in her hand. With information synchronized, she would be able to give good clues on her next turn should she have the opportunity. 

Additionally, playing r2 gives no opportunity to do priority finesses. According to your understanding, if she plays r2 nothing happens. If she plays unknown 2 it causes a prio bluff. 

By having unknown cards have priority, if she plays r2 it causes a prio finesse or a prio bluff. 
I see only benefits to making cards that are unknown have higher priority. You lose nothing, you gain prio finesses, and better synchronization of information.

Zamiell:
> You lose nothing

LOL

sounds exactly like me when i invented priority

EhDrone:
> According to your understanding, if she plays r2 nothing happens. If she plays unknown 2 it causes a prio bluff
This is wrong - it is directly defined that playing an unknown card never triggers priority: "If one of the cards is not completely known yet (e.g. a playable 2 of unknown color), then the player might want to play that card first in order to find out exactly what it is. In general, nothing special is triggered by players playing an unknown card."

Second - conceptually you are overvaluing playing an unknown card. In my example, the other 2 lead absolutely nowhere regardless of which one it is (just like the known 2). So is there a minor benefit to knowing WHICH 2 it is? Sure - but it doesn't actually matter because it will still play the following turn, and regardless of which 2 it was it doesn't change anything about the board state.

What you LOSE by making a strict rule that every unknown card has priority over a known card is the freedom to allow someone to choose based on context. One quick example I can come up with is that Alice sees red 4 near someone's chop with possible a critical or near-playable unique card off-chop. If she plays red 2 and draws red 3, we can save red 4 and keep the off-chop card around an extra turn without needing to clue it, as red 4 can play now. If she plays the unknown 2 and draws red 3, now the red suit is delayed playing the 4 and maybe Cathy (or whoever) needs to discard in the interim, potentially losing a unique 3 (or another 4) that is on the bottom of the deck.

EhDrone:
You can open an issue for sure for more discussion - again to clarify, this PR is not a convention change but a clarification for how the conventions exist because the detail was missing in the Priority chart and contained only in the flow chart.

Zamiell:
see my edits

PR #1201, opened 2024-04-29
Update level-3.mdx (Early Fix Clue)

topica314:
Cards should be fixed as soon as possible, regardless of whether the misplay happens immediately or not.


Zamiell:
thanks

PR #1207, opened 2024-05-25
Adding hard bubblegum bluffs

### LINKED ISSUE #1191: Allowing The Bubblegum Bluff (Two-Away-From-Playable Form) to bluff pink cards as a hard bluff
Pros:
- Efficient enough to satisfy MCVP and some extra.
- More common than a double finesse or a half bluff.
- By the definition gets cards blind played from a slot that is not the finesse position which are more difficult to get.
- It works as a hard bluff because the clue receiver knows the exact identity of their pink card.

Cons:
- More intuitive. There is a reason why they are called hard bluffs.
- Removes the more efficient possibilities.
- An another reason that I did not think of.

Stuurrm:
Closes #1191


Stuurrm:
Will note that forgot to mention the double bluffs in the issue but it would be wierd and confusing to allow hard bluffs for one of them but leave the other out.

PR #1216, opened 2024-07-22
Clarify Priority with Known+Unknown cards

english5040:


Zamiell:
I'll have ADrone take a look at this PR, since he was in charge of the last priority change/clarification

EhDrone:
I might be missing something - but you updated it to describe the opposite scenario of what the section is trying to explain. Sorry if I am reading it wrong -- but the point of that section is to let people know that playing an Unknown card instead of the known card, but you updated the section to describe playing a known card over an unknown card.

english5040:
There are two cases the original was trying to describe: when does playing an unknown over a known cause a PF and when does playing known over unknown cause a PF. In the first case that generally does nothing, and the second case is the important one to discuss, wherein the orig. doc was confusingly phrased

english5040:
The orig. doc was trying to say that the known has to have lower priority than all unknown superpositions but said it badly and reversed it

english5040:
additionally, according to the priority table, the original examples were invalid, so I changed them to reflect the current priority table

english5040:
hm, I can rephrase some of the wording, I think

english5040:
there we go, now it's all about whether the known has strictly lower priority than the unknown identities

english5040:
This PR is now obsolete after a discussion with ADrone; it reflected a common interpretation of the priority table which was not the intended meaning

Issue #1212, opened 2024-07-22
The Priority with Unknown Cards section remains confusing and inaccurate


english5040:
Alice didn't even play an unknown card in the example, so the description of this section doesn't even make sense. In addition the example is inaccurate because unknown cards have priority over leftmost (criterion 6) if they're tied in all criteria above 6.

english5040:
making a pr

Zamiell:
should i close this now?

Zamiell:
also i removed [this flowchart](https://raw.githubusercontent.com/hanabi/hanabi.github.io/main/static/img/flowcharts/priority-unknown-flowchart.png) from the doc just now since it fails to capture the following scenario:

- Alice has a globally known red 4 clued in her hand.
- Alice has a `blue 2, green 2`  clued in her hand.
- Bob has a globally known red 5 clued in his hand.

Here, Alice is expected to play the red 4 into the red 5. Thus, the rule of "playing unknown cards never cause a Priority Finesse" doesn't actually apply here. I don't really feel like updating the flowchart, so I'll delete it for now, if someone wants to recreate it then go for it.

Zamiell:
(the situation i sketch in my previous comment above is actually just a repeat of the example that is already in the doc under Trust Finesse).

for reference the deleted flowchart is here: https://github.com/hanabi/hanabi.github.io/blob/3b312e95131c1c130fd885d3e31982b131cddfc1/static/img/flowcharts/priority-unknown-flowchart.png

english5040:
yeah I'm closing it. I'll make another issue if I want to discuss adding "unknown identity" as a part of the table.

I am of the opinion that priority was so confusing, we should just stick with the current convention, hopefully the PR'd priority table should really cut down on confusion

Zamiell:
> I am of the opinion that priority was so confusing

yes priority is basically the worst convention that we have, whoever invented it is an idiot

PR #1218, opened 2024-07-24
Rewrite priority table and associated sections to reflect their intended meaning

english5040:


Zamiell:
this should probably wait until #1216 is merged

Zamiell:
nevermind, looks like that PR is closed now

@EhDrone do you wanna review this

EhDrone:
You swapped the order of examples to be positive then negative, that is fine I don't mind. In the second example though, I would suggest adding the step of analyzing the "rainbow 1" as well and not just the higher priority item to give the full train of thought and to explicitly show that even though one option had higher priority, the other did not - and therefore no priority plays are called for.

I am not a pro in github - is there a way to sample what the site page will actually look like based on the PR here? I am just curious what it will look like formatting wise with the 'Unknown Card' row from the priority table pulled up to the top.

Zamiell:
> is there a way to sample what the site page will actually look like based on the PR here?

i could make CI publish a test build, but its kind of complicated.

you can build it on your local computer though by checking out the branch.
if you have the github cli installed you can just type `gh pr checkout 1218`.
i can walk you through it if you want sometime

J-Fricke:
Does this account for playing a known 1 with unknown ones (or similar situation) creating a priority play as unknown in and of itself is a higher priority? I think it should and we recently used this in a game.

english5040:
That isn't how priority is intended to work. It does seem like this is how most people interpret the current priority table- if you also think so, please make an issue for discussion, I've exhausted myself by now.

english5040:
A Discord thread was made for Priority discussion related to this

EhDrone:
I would suggest updating the table text for unknown to specifically say "Apply the table to each possible identity of the unknown card: see [here](#priority-with-both-known-and-unknown-cards). (**Reminder**: playing an unknown card generally doesn't trigger priority except in special cases _Priority_.)

Also - I guess I didn't read the doc carefully enough myself to begin with, as there is this section already - and goes to show that people playing without this understanding just didn't read the priority section entirely to begin with, just looked at the chart and moved on assuming they understood everything: https://hanabi.github.io/level-20/#the-priority-finesse-special-case

We should either delete the section I wrote, and rename this one to be more descriptive, or delete this section in favor of the section I wrote / is updated in this PR.

Zamiell:
> We should either delete the section I wrote, and rename this one to be more descriptive, or delete this section in favor of the section I wrote / is updated in this PR.

do you want to do that in a separate pr, or make english do that in this pr?

english5040:
ADrone is going to be inactive for a while; can Zamiel review this?

jameelahuq (changes requested):
Flow chart uses the word "dumb" which is difficult to parse as an action step. Also it gives permission for players to say that each other's moves were "dumb" which isn't the best feeling in the world. 

Also, if was more clear that this was for the...uh...whatever finesse it's supposed to before (I don't remember what finesse James said it was for) then that would be helpful.

Zamiell:
> Does this account for playing a known 1 with unknown ones (or similar situation) creating a priority play as unknown in and of itself is a higher priority? I think it should and we recently used this in a game.

i'm glad it worked in the game but afaik that specific pattern is undefined, adrone can correct me if i am wrong

Zamiell:
PR looks great to me, thanks very much english.
(what is your name on the Hanabi website?)
i made some small edits that you should probably review now.
when @EhDrone gets off of vacation, he should review this PR as well.

Zamiell:
> Flow chart uses the word "dumb" which is difficult to parse as an action step.

see https://github.com/hanabi/hanabi.github.io/issues/1212#issuecomment-2265976914

english5040:
I'm ejwu/ejwu2 on discord. I like the new changes, they definitely make things more concise

PR #1223, opened 2024-08-04
Reorganize Reversed conventions by level

bnordli:
In the reversed conventions, I noticed that "No Order Chop Moves" was listed after "The 1's Fix Clue", which doesn't make any sense. 

So I decided to try to organize the conventions by level, just like the Pink conventions.

(In Pink, the "Turnabout Ejection" is listed as Level 21, but here I decided to group it with the "1 Color Ejection" at level 16. I am open for discussion around this..)


Zamiell:
seems fine thanks

PR #1224, opened 2024-08-05
Revise convention docs for grammar and style

english5040:
This is a full edit of the convention docs for grammar, style, and consistency in formatting.

(currently incomplete)


english5040:
I see a lot of "we agree that...", "should..." and other similar phrases, these can be made more concise. Given that this is a conventions document, I prefer to say "is" in these cases, even though conventions are not truths but agreements.

Zamiell:
ok this was an absolute nightmare to review, i will close any other big prs like this in the future, please open many concurrent prs for edits of this nature

PR #1226, opened 2024-08-05
Patch 4

Gingy120-github:
Bob sees that Alice didn't clue Donald's b2, but he doesn't know whether his hand contains the other b2. So, he cannot know for sure that this is a 5 Pull Double Finesse. Instead, he looks to the Visible Rule and will assume that this is just a 5 Stall. To make this more clear in the documentation, I suggest changing Donald's b2 to a clearly gettable card like y1.


Zamiell:
thanks

Issue #1220, opened 2024-07-27
Remove double tempo clues


ricardodd2:
I beleive single (left) focus tempo clues, potentially tccm, have more benefits than the double tempo clue conventions. That's because the majority of the time the team tempoes a card is to protect the clue reciver's chop and is common enough that a clue touching a single card is bocked, especially in variants: white (no color on white cards), ranbow (blocks color), pink (blocks rank clues), ...

The convention change also feel to me to fall more in line with the rest of the H-group conventions and clears up possible confusions when a tempo clue touching multiple cards is given:
- every other clue in H-group has a single focus and no promises are usually made on cards outside the focus
- when a player receives a clue on multiple cards the meaning can only be one, so tempo clues become easier to interpret. e.g. b2 i1 played on the stack. Bob has two 3s and the leftomst is promised 3 from a previous 3 bluff. Alice gives a rank 3 clue touching Bob's 3s. Is Bob supposed to play only b3 from the left or both b3 and the other 3 as i2? I personally don't know the answer with current conventions, but with the removal of double tempo clue, only one interpretation remains.

Is there a way to put a voting system? Edit: I guess I'll leave the usual reaction (thumbs up/down) as a voting system

Zamiell:
> Is there a way to put a voting system? 

a voting system isn't super helpful because each person's vote holds a custom amount of weight; see: https://github.com/hanabi/hanabi.github.io/blob/main/misc/convention-changes.md

in general, i would prefer that people actually take the time to put a thoughtful comment on an issue instead of the lower-effort option of voting / slapping an emoji on

gw12346:
I think as the conventions stand, there are a number of situations where some players treat reclues as single tempo clues even though the conventions say that those reclues are double tempo clues. 

The justification given for two cards not being both playable, left-to-right, is generally "context". 

This is more often in variants involving rainbow or pink suits, so I think we could benefit from more clarification in the docs, which only talk about the novar case.

english5040:
There's a lot of fuzziness around double tempo clues. Instead of going through the work to define them, removing them altogether seems pretty reasonable to me.

Zamiell:
double tempo clues are not common, so i am for this proposal.

PR #1228, opened 2024-08-21
Remove double tempo clues

gw12346:
According to consensus in #1220.


Zamiell:
imo add a case to the TCCM section that outlines the situation where red retouches 2 cards, but only the left most one is treated as playable and the other card is chop moved

Zamiell:
actually can you do that in a separate PR? im going to announce this now

PR #1229, opened 2024-08-22
Remove Double Tempo references

english5040:


english5040:
Looking at the Tempo Clue section, the red 2-5 tempo clue looks a bit redundant now. I think the example is important, but now it's important because it's an example where the focus of the tempo clue is a bit flexible. I know that James stated on Discord that the focus of a tempo clue can be flexible, but I'm not sure how to phrase it in the doc yet.

Zamiell:
awesome thank you

Zamiell:
> the red 2-5 tempo clue looks a bit redundant now

ill remove it

PR #1236, opened 2024-09-17
Add details about information resets after a bomb

3geek14:
We reset our assumptions about cards, but we don't reset game state information. If a card was touched by a mistaken clue, we remember that the card was touched when we later think about the focus of new clues.


Zamiell:
thanks

PR #1250, opened 2024-10-08
realign trash moves in pink 

rz-1:
convention proposal: https://discord.com/channels/140016142600241152/1289284121260068937/1289284121260068937

- Now all cards touched in trash cm and trash push are trash. This realigns these trash moves with trash bluff/finesse in pink.
- omni now inherits the trash moves conventions from pink (removed trash push section)


Zamiell:
thanks

PR #1252, opened 2024-10-11
Update level-8.mdx

newduke:
Added an endgame heuristic


Zamiell:
i dont know if this heuristic is correct, and furthermore im not sure if we want to give specific advice on end-games, since they can vary.

newduke:
It's a guide, not an absolute truth.  Beginners have a hard time making these choices, so may as well give some general advice.

Zamiell:
ok, given that no-one else has commented, i will close this PR for now.

PR #1254, opened 2024-10-27
fix: clarify 5s Chop Move

3geek14:
The text currently says that a clue is only a 5s Chop Move if the chop is trash. The player receiving the clue is the only one who doesn't know whether their chop is trash, so they can't use that information to interpret the clue.

Remove the requirement for chop to be trash. At level 4, all 5s clues that newly touch a one-before-chop 5 are interpreted as 5s Chop Moves. Players won't do this when the chop is trash.

This eliminates the potential for Alice to give a 5s Play Clue to Claire, finessing Bob. From discussion on Discord, it sounds like players are likely to misinterpret this clue, particularly at level 4: https://discord.com/channels/140016142600241152/1299966501343723540


Zamiell:
see my edits

Zamiell:
also, her name is Cathy, not Claire.

3geek14:
Edits LGTM. And thanks for the name correction.

PR #1257, opened 2024-11-01
Move clarity principle to level 10

downvoid:


Zamiell:
i thoguht we wanted to move it to level 5 (?)

downvoid:
you said level 5 is already too long, so level 10 was the next level that fits well

downvoid:
oh wait, I misread that

Zamiell:
level 6, not level 5

downvoid:
there, it's now at level 6

PR #1258, opened 2024-11-01
add the word chop

downvoid:


Zamiell:
you have to update level 11 and the beginners guide too, right?
also, it is more gramatically correct to say "their chop card", not just "chop".

downvoid:
wait, level 11? what do I have to change there?

downvoid:
I've fixed the other stuff, but I don't see anything on level 11 related to this

Zamiell:
level 8 i meant, not level 11

downvoid:
ah there we go

Zamiell:
the grammar is still weird, i would word it like this:

"the period of time before any player has discarded their chop card"

downvoid:
I just copied exactly what you put

Zamiell:
you only updated 1 page though, all 3 pages should have the same wording.

downvoid:
is this what you wanted

Zamiell:
yep

PR #1259, opened 2024-11-01
Discuss white/gray clue arrows in beginner guide

Gingy120-github:
Added a small section in "Card Layout" to introduce the gray arrow for retouches. Also added a small section in "Single Card Focus" explaining new=white and old=gray if retouched.

### DIFF
--- a/docs/beginner/card-layout-2.yml
+++ b/docs/beginner/card-layout-2.yml
@@ -15,6 +15,12 @@ players:
   - cards:
       - type: x
       - type: x
+      - type: x
+      - type: x
+      - type: x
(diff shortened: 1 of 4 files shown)


Gingy120-github:
I remember a game I played where there was a beginner who thought that the entire hand became clued when any clue was given (because some cards are explicitly labeled with positive information while the other cards are quietly labeled with negative information). I also remember a time when a beginner was confused about new vs old (perhaps due to not reading the guide closely). So, I suggest these changes to help prevent these misconceptions.

Zamiell:
seems fine, thanks

PR #1260, opened 2024-11-01
Patch 9

Gingy120-github:
Fixes a grammar error in level 14.

Adds a new section to level 18 to clarify the interaction between trash moves and elimination. Added some new images too.


Gingy120-github:
Sorry, the many commits that I made while editing the same file over and over again seem to have stayed and not collapsed into a single commit. Not sure how to condense it all, unfortunately.

Gingy120-github:
opening new pull request with fixes to tests

Zamiell:
it is not required that PRs be a single commit, as we will squash them all when the PR is merged.
thus, feel free to make a PR with as many commits as you like, it is irrelevant.

Zamiell:
however, a PR should do 1 thing, so if you are fixing a grammar mistake, do that in a separate PR from adding a new section.

PR #1263, opened 2024-11-01
Update lvl 18 with clarification on trash moves

Gingy120-github:
Adds a new section to lvl 18 to clarify what already happens with Trash Chop Move (I guess this one is also "assisted"...) when missing card is already in someone's hand and Trash Bluff.

The Trash Chop Move triggers elimination note writing, while the Trash Bluff does not.


Zamiell:
i don't understand the example, bob would just bomb the purple card as p1

Zamiell:
the stacks should be listed from left to right. also you forgot to update the other example (?)

Gingy120-github:
Lemme take a closer look; things must have gotten a bit scrambled when I was editing so many things at once.

Gingy120-github:
The purple bomb is a typo; red 4 on stacks is supposed to be purple 4. When I fix everything, do I close this and open a new pull or is there a way to update this pull?

Zamiell:
just update the existing pull request by making commits to the branch that you opened the pull request with.
(the pull request is just in sync with your branch.)

Gingy120-github:
Ok, I believe I've fixed all the errors. I can't commit the file directly though, since you've made some edits. Would you like me to post my version somewhere so you can compare, or are you already working on fixing everything and don't need it?

Zamiell:
you can type `git pull` to pull down my edits i think

Zamiell:
(use `git config --global pull.rebase true` to prevent merge commits)

Gingy120-github:
Unfortunately I ~~didn't pay enough attention in my college computer science courses~~ am not able to get my branch open in VSCode to type in the command, so I'll just wait for you to finish making changes before I go in to suggest/clean things. Based on what I can see so far, though, I really like the separation of the forward/reverse.

Zamiell:
im done making edits now.

in the future, always use red as the primary color, and then blue if you need an additional color, and then green if you need a 3rd color.

Zamiell:
my next concern is that the section called "Interaction Between Elimination & Potentially Unknown Trash (Reverse)" probably does not belong here.

the primary point of that section is to communicate the idea of "Bob knows that for the purposes of playing into _Trash Bluffs_, you are not supposed to assume that the unknown cards in your hand match the trash."

thus shouldn't it be in the Trash Bluff section?

Gingy120-github:
Hm, that's a good concern. I think that section has merit being in level 14 to communicate the idea of "don't assume that Alice is asking you to play your 5 to show Cathy that has red card is trash", but at the same time it also has merit being in level 18 to communicate the similar but distinct "don't write elimination notes on your hand even if you blindplay finesse position".

Noted about the primary colors, btw. I did not know that, but that convention keeps things cleaner. Will do, moving forward

Gingy120-github:
I think we have a major problem. Someone brought up to me a very good question concerning the "Interaction Between Elimination & Potentially Unknown Trash (Forward)" and the Reverse Trash Finesse.

I'll use the same example board as the one under the Reverse Trash Finesse section to explain.

- Stacks: r1 y2 g2 b2 p2
- Alice: clues number 2 to Bob
- Bob: g[2] xx xx xx xx
- Cathy: r2 xx xx xx xx

At level 14, Bob cannot know whether Alice is doing Reverse Trash Finesse or just direct Play Clue on r2, so he discards to cover both cases. Regardless of what happens, everyone marks cm on Bob's cards, and Cathy blindplays finesse.

However, the proposed clarification seems to conflict with what is written in level 14. Let's update the example slightly.

- Stacks: r1 y2 g2 b2 p2
- Alice: clues number 2 to Bob
- Bob: g[2] xx xx xx xx
- Cathy: r2 r2 xx xx xx

Now Bob knows for sure that his 2 is trash. Like before, he discards. Everyone then marks cm on Bob's cards. But how does Cathy know whether this is a Reverse Trash Finesse or just a Trash Chop Move? 

The proposed clarification says that Cathy will just mark r2e and then do nothing else. Perhaps a discussion about Reverse Trash Finesse vs Trash Chop Move (assisted) is needed, or we just discard that section of this proposal?

Gingy120-github:
I'm willing to start a discussion of replacing Reverse Trash Finesse with the Trash Chop Move interpretation, since I don't think I've ever seen Reverse Trash Finesse be used, but of course I am a newer member so probably don't know the value of it.

Gingy120-github:
Actually, I think this is resolved if Bob has a clue to spend. 

- Stacks: r1 y2 g2 b2 p2
- Alice: clues number 2 to Bob
- Bob: g[2][kt] xx xx xx xx
- Bob: clues number 2 to Cathy
- Cathy: xx r[2] r[2][kt] xx xx

Now Cathy is playing r2 and everyone is in the know.

downvoid:
> I think we have a major problem. Someone brought up to me a very good question concerning the "Interaction Between Elimination & Potentially Unknown Trash (Forward)" and the Reverse Trash Finesse.
> 
> I'll use the same example board as the one under the Reverse Trash Finesse section to explain.
> 
>     * Stacks: r1 y2 g2 b2 p2
> 
>     * Alice: clues number 2 to Bob
> 
>     * Bob: g[2] xx xx xx xx
> 
>     * Cathy: r2 xx xx xx xx
> 
> 
> At level 14, Bob cannot know whether Alice is doing Reverse Trash Finesse or just direct Play Clue on r2, so he discards to cover both cases. Regardless of what happens, everyone marks cm on Bob's cards, and Cathy blindplays finesse.
> 
> However, the proposed clarification seems to conflict with what is written in level 14. Let's update the example slightly.
> 
>     * Stacks: r1 y2 g2 b2 p2
> 
>     * Alice: clues number 2 to Bob
> 
>     * Bob: g[2] xx xx xx xx
> 
>     * Cathy: xx r2 r2 xx xx
> 
> 
> Now Bob knows for sure that his 2 is trash. Like before, he discards. Everyone then marks cm on Bob's cards. But how does Cathy know whether this is a Reverse Trash Finesse or just a Trash Chop Move?
> 
> The proposed clarification says that Cathy will just mark r2e and then do nothing else. Perhaps a discussion about Reverse Trash Finesse vs Trash Chop Move (assisted) is needed, or we just discard that section of this proposal?

this is just a normal reverse trash finesse

Gingy120-github:
Yes, I know. Cathy questioning which of the two cases is correct is highlighting that this proposal is conflicting/rewriting existing conventions and so we either need to drop the proposal or discuss changing the convention

Zamiell:
> But how does Cathy know whether this is a Reverse Trash Finesse or just a Trash Chop Move?

It's the same algorithm as the what is listed in the TTE section:

> When this occurs, and there is exactly one "missing" card, players should write an Elimination Note in their hand for the missing card. This is called Trash Touch Elimination.

Since there are two red 2's missing, then it requires a blind-play.
If you modified your hypothetical slightly by putting one copy of red 2 in the trash, then Cathy would write TTE notes, because it would only be one card instead of two.

Zamiell:
for now ill get rid of the second section

PR #1267, opened 2024-11-02
Update level-9.mdx

Gingy120-github:
Of the precedence 4 stalling moves, fill-in clues have the fewest restrictions because they are allowed in DDA and Locked Hand severities.

Meanwhile, locked hand saves and 8 clue saves are only allowed under more extreme circumstances.

Thus, it would make the table more clean if the precedence 4 clues were sorted by severity allowability.


Zamiell:
Ok

PR #1268, opened 2024-11-02
Update level-9.mdx

Gingy120-github:
Clarifies when FPE is allowed, based on my understanding of it.

I had a few games where it was thought that FPE can be done in any situation where there is only one Play Clue left to give and it would possibly be killing someone's Finesse. I thought this myself, once, until I got questioned heavily in review lol.

However it seems that FPE is only meant to be used in the case that the user is already in a Stalling Situation. If true, I suggest we add something saying that FPE only applies when already in a Stalling Situation and cannot be used to create one from thin air.


Zamiell:
see my edits

PR #1271, opened 2024-11-16
Update level-12.mdx to add something about optional prio.

gw12346:
This PR is for @Zamiell, discussion at https://discord.com/channels/140016142600241152/1299174627653718026/1306578329842880583.

### DIFF
--- a/docs/level-12.mdx
+++ b/docs/level-12.mdx
@@ -15,75 +15,6 @@ import CluingOnes1 from "./level-12/cluing-ones-1.yml";
 import CluingOnes2 from "./level-12/cluing-ones-2.yml";
 import CluingOnes3 from "./level-12/cluing-ones-3.yml";
 
-## General Principles
-
-### Context
-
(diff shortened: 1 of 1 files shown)


Zamiell:
can you explain why you are moving the "General Principles" section in this PR?

gw12346:
In my experience, players learning level 12 are great at learning conventions (focus inversion and stale 1s) but some seemingly skim over "General Principles": stealing clues when they have multiple cards to play, unnecessarily saving cards they themselves hold, and giving clues other players can't build off of.

We're great at teaching players _how_ to clue, but I wonder if we need more emphasis on _whether_ to clue.

Zamiell:
all of the levels follow the same ordering, with general principles being at the end.
i dont think you can change the ordering in just one level and keep the ordering in all of the other levels the same, that would be a bug.
if you want to do a major change like that i think you first need to open an issue, for now revert the reordering.
in general, please only have your PRs be doing 1 thing at a time.

Zamiell:
from the "files changed" tab it looks like it wasn't restored to original order

Zamiell:
see my updates, thanks

PR #1272, opened 2024-11-16
Level 3: Add fresh 1s example

3geek14:
Add an example for playing unknown 1s when multiple new 1s are in the player's hand.

This was inspired by a game I just played where 2 players marked an ocm, but the player who played a 1 thought they were playing in the expected order and didn't mark it.

The first commit adds a new example numbered 2a between examples 2 and 3. The second commit renames 2a -> 3 and 3 -> 4.


3geek14:
If this were Mercurial rather than Git, I would've been able to track the filename changes, so I would've done it all in one commit. Ah well.

Zamiell:
> The first commit adds a new example numbered 2a between examples 2 and 3. The second commit renames 2a -> 3 and 3 -> 4.

no need to document what the individual commits do, i typically review PRs by just looking at the "files" tab in the PR, which shows the collective changes from all the commits in the PR

Zamiell:
> If this were Mercurial rather than Git, I would've been able to track the filename changes, so I would've done it all in one commit. Ah well.

i dont think that matters, all PRs are squashed, so it doesn't matter if your PR has 1 commit or 100 commits, they all just get squashed into 1 commit anyway, and yes git will track the changes between renamed files automatically.

3geek14:
Git will track that `docs/level-3/playing-multiple-1s-3.yml` has been edited and that `docs/level-3/playing-multiple-1s-4.yml` is brand new. Mercurial is able to track that the old `docs/level-3/playing-multiple-1s-3.yml` was renamed to `docs/level-3/playing-multiple-1s-4.yml`, and that there's a brand new file named `docs/level-3/playing-multiple-1s-3.yml`. I think that distinction can make code review easier, which is why I split it into two commits. (But, however you're used to reviewing PRs will probably continue to be easiest for you.)

I like your additions to the PR.

Zamiell:
> I think that distinction can make code review easier, which is why I split it into two commits.

yeah, it would if the PR reviewer was reviewing individual commits, but I think the normal thing that people do on GitHub is to click on the "files changed" tab and review the entire PR at once.
the REASON that this is the normal workflow is because it is common for people to add things that are unnecessary and then later on remove them in the final "solution".
so, if someone was reviewing the individual commits, then they would see a bunch of code and waste time reading through it, when all that code would just be deleted again in the next commit.
hopefully that makes sense.

3geek14:
> the REASON that this is the normal workflow is because it is common for people to add things that are unnecessary and then later on remove them in the final "solution".

I agree it's common for people to add unnecessary code in one commit, remove it in a later commit, and upload all of those commits in the PR. I have a long rant about that, which I won't assume you're curious to hear. 😉

Zamiell:
assuming that you squash commits (which is considered best practice AFAIK), then all the commits are going to get deleted anyway.
for that reason, it is unnecessary busywork to make the PR submitter clean up their commits before submitting the PR.
doing that would arguably be worse too in the sense that the PR reviewer would not get to see the individual things that the PR author tried when converging upon the final solution.

Zamiell:
regarding the PR, I think the provided example situation is rare enough that the value added from the example is outweighed by the clutter it would add to have it in.
(it is a design principle of the doc that we want to keep it as short as possible.)

Zamiell:
can you remove the "playing-multiple-1s-4.yml" file and also explain what the changes to the "playing-multiple-1s-3.yml
" file are doing?

3geek14:
Wouldn't it make more sense to remove the example I added (3) and keep the old one (4, renumbering it back to 3)?

Zamiell:
I copied the new text from this PR into the doc.
see this commit: https://github.com/hanabi/hanabi.github.io/commit/f8c735a884416bf32f76505408b18ea888f625ef

i'll close this PR for now, but thank you for taking the time to add an example and try to make the doc better, the effort is not lost on me.
in the future maybe a better course of action would be to ask if a certain addition would be okay, that way it potentially avoids wasting work.

Zamiell:
> However, if you're saying that it's more work for you to review a PR,

nah, not really, was just trying to save you some work =p
feel free to do whichever is less work for you.

PR #1274, opened 2024-11-18
added early game info to level 4: chop moved playable cards

hm1298:
per discord discussion, early game requires play clues be given to chop moved cards when possible


Zamiell:
thx

PR #1277, opened 2024-11-24
Remove reference to Context in Clarity Principle

RandomGuyJCI:
Since Clarity Principle has been moved to Level 6, it seems that linking to a section on a later level is unnecessary, especially since the text still makes sense without the link.


Zamiell:
thanks, i missed that

PR #1284, opened 2024-12-13
Update level-9.mdx - Defining Stalling Situation

HungryBradbury:
I have always felt this level does a poor job explaining what a stalling situation is, given that this level has everything to do with Stalling. So I've added a section defining a stalling situation.


Zamiell:
see my edits

Zamiell:
can you do another pr for that

PR #1285, opened 2024-12-16
Update dark-pink.mdx

newduke:
5NE has lower precedence than turnabout, thus it cannot be used


Zamiell:
Can you make it pass ci

PR #1290, opened 2024-12-21
Add colours to the yes/no nodes for clue focus

3geek14:
This commit demonstrates how to set node background colours from outside the flowchart definitions, with minimal extra code needed within the flowchart itself.

Inside the flowchart, we need a line for each CSS class we want to add to nodes within the chart. In the global CSS file, we define colours as variables and apply those colours.

This is intended as a proof-of-concept, rather than as a final version. I have not put much thought into whether the colours are good; I merely wanted them to be clearly present and based on the user's light/dark preference.

image

image

### DIFF
--- a/docs/level-1.mdx
+++ b/docs/level-1.mdx
@@ -73,6 +73,9 @@ D3N(No)
 D3YL([The focus is on the chop card.])
 D3NL([The focus is on the left-most card.])
 
+class D1Y,D2Y,D3Y node-yes
+class D1N,D2N,D3N node-no
+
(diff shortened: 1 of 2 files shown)


Zamiell:
can you remove the changes to package.json and package lock? i dont think they are necessary, they look like a mistake on your end

3geek14:
Yep, those must've gotten messed up when I was tying to update node/npm in order to build locally.

Zamiell:
great work thank you

PR #1291, opened 2024-12-21
Formatting in Lvl 1 and Lvl 19 Flowchart

HungryBradbury:
- Interchanged the node shape between Yes/No nodes and Action Nodes, because while working on lvl 7 flowchart I found that the stadium shape of action nodes can't handle large text.
- Included class def for Yes/No nodes so that they change color with light/dark theme.
- Formatted some text in the nodes.

### DIFF
--- a/docs/level-1.mdx
+++ b/docs/level-1.mdx
@@ -58,20 +58,20 @@ This process is represented in the following flowchart:
 ```mermaid
 flowchart TD;
 
-S([A clue is given.<br />What card is focused?])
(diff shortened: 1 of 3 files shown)


HungryBradbury:
Okay, didn't realize if I pushed another commit while my PR was open it would add it to the PR.

So new addition - Level 7 Mermaid Flowchart

Zamiell:
> Okay, didn't realize if I pushed another commit while my PR was open it would add it to the PR.

yes, you need to start a new branch for that

PR #1295, opened 2024-12-23
No Play Clues with Number 5 in Early Game

Gingy120-github:
updated level 1 and level 2 according to https://discord.com/channels/140016142600241152/1307505935450636309

### DIFF
--- a/docs/level-1.mdx
+++ b/docs/level-1.mdx
@@ -98,7 +98,7 @@ D3N --> D3NL;
 - The _Early Game_ is the period of time before any player has discarded their chop card.
 - Players **must** "extinguish" all of the available _Play Clues_ and _Save Clues_ on the board before ending the _Early Game_.
   - As outlined in the [beginner's guide](beginner/minimum-clue-value-principle.mdx), "extinguishing" all the _Play Clues_ does **not** include giving _Tempo Clues_. _Tempo Clues_ do not meet _Minimum Clue Value Principle_, so in general, they should not be given.
-- For level 9 players, there are [some additional rules](level-9.mdx#the-early-game-severity-1-stalling) about the Early Game. (Don't worry about them for now if you are only level 1.)
+- There are some additional rules about the Early Game for [level 2 players](level-2.mdx#the-5-stall-cluing-off-chop-5s) and for [level 9 players](level-9.mdx#the-early-game-severity-1-stalling). Don't worry about them for now if you are only playing at level 1.
 
 ## Special Moves
 
--- a/docs/level-2.mdx
+++ b/docs/level-2.mdx
@@ -17,12 +17,25 @@ import WrongPrompt from "./level-2/wrong-prompt.yml";
 
 ### The 5 Stall (Cluing Off Chop 5's)
 
-- Normally, you are only allowed to give a _Save Clue_ to a 5 if it is on chop. So if you use number 5 to clue a 5 that is **not** on chop, then it will typically look like a _Play Clue_ on that 5 (or some other advanced strategy).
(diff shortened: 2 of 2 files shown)


HungryBradbury:
Can you also do one that prohibits Early 5 Chop Move in level 4 and level 9? While I'm not in favour of that, it'll be better for the Convention Guide to reflect how H-Group plays.

Gingy120-github:
> Can you also do one that prohibits Early 5 Chop Move in level 4 and level 9? While I'm not in favour of that, it'll be better for the Convention Guide to reflect how H-Group plays.

I can, but I would rather have the group finish the discussion about moving e5cm before writing anything.

Zamiell:
sorry for only getting to this PR now.
anyways, the problem i see with this PR is that it does not actually map on to low score phase.
meaning that it is currently possible for it to be not the low score phase and still in the early game, and now this convention actually changes the way we play rather than acting as a clarification for lower levels.
since low score phase does not map on clearly to the early game, i don't see a good solution to this problem, since we certainly don't want to introduce the concept of the low score phase at level 2.
and we equally don't want to delete the level 2 convention at level 19 (e.g. "just kidding about that level 2 convention, it actually means low score phase instead of early game")

Zamiell:
here's my counterproposal:

- the text about "nothing else to do" is deleted from level 2 and moved to level 9.
- thus, 5 stalls are now legal in early game always, regardless of card/game configuration.

Gingy120-github:
Yeah, I was aware about the distinction between _Early Game_ and _Low Score Phase_ and did understand that it's possible for the game to still be _Early_ but also _Normal Score_. The purpose of this PR was to fix the problem where a player, especially below level 9, would give a rank 5 clue while having something else to do, expecting the team to understand it as a _Play Clue_... but then it's assumed to be a mistake so nothing happens. This is a problem because the convention document doesn't match the way the group plays. The level 2 conventions say that _5 Stalls_ are only legal when there is nothing else to be done, so since there IS something else to be done, it must not be a _5 Stall_ after all.

So, this PR truly did aim to update the conventions such that they are in line with how the group behaves (which I guess, in a way, is changing the way we play), and I believe it has the following effects during the _Early Game_:

- Below level 9, a rank 5 clue to an off-chop 5 does nothing
- At or above level 9 and before level 19, it still does nothing
- At or above level 19, a rank 5 clue does nothing unless:
  - _Early 5 Chop Move_
  - _5 Pull or _5 Number Ejection_ or _5 Number Discharge_
- And finally, at or above level 19, if it is _Early Game_ but _Normal Score_, a rank 5 clue does nothing again... which, now that I think about it, is kinda strange.

It feels like level 19 is a checkpoint after which players are expected to understand the purpose behind rank 5 clues and their timing, so I see what you mean by "just kidding about that level 2 convention, it actually means low score phase instead of early game". Part of me still feels like there is a line and that it needs to be drawn somewhere, though, and that line is caused precisely because of the non-mapping you mention.

**Anyway, I like your counterproposal.** I am slightly worried that it doesn't emphasize at level 2 that there is no reason for a player to perform a _5 Stall_ if they have something to play, but perhaps we can trust the reader to make that understanding. Should we then remove all the text in this PR about "known mistake", or just move it to level 9?

Also I forgot to post the Discord thread for this proposal so [here it is](https://discord.com/channels/140016142600241152/1289276242495017061/threads/1311621985683640320).

Gingy120-github:
I guess the entire thing about "the line having to be drawn" is resolved if everyone is willing to enforce illegal _5 Stalls_ and treat them as _Play Clues_, especially since at level 19 (during Early & during Normal) it appears players already do.

But I don't think players are willing to do that in general, because it is valuable to give leniency to lower level players.

Zamiell:
ok if noone else has any input i will implement my counterproposal in a separate commit

Zamiell:
also https://github.com/hanabi/hanabi.github.io/commit/154f5a4ea0ae423318da8b9e875921b03e01216a

PR #1296, opened 2024-12-24
No Reverse Trash Finesse on Critical Cards (clarification)

Gingy120-github:
updates lvl 14 to clarify why no RTF on critical cards. should heal the interpretation problem between lvl 14 and lvl 18.


Zamiell:
see my edits

PR #1297, opened 2024-12-28
Formalizing Known-Trash Discard Order with Trash from UTD/UDD and other minor tweaks

HungryBradbury:
This commit:
- Formalizes Known Trash Discard Order in Lvl 14.
- Highlights the convention of discarding trash from UTD/UDD before other known-trash cards in Lvl 14 (which I feel isn't visible enough) as well as integrates it into a revised Known-Trash Discard Order at Lvl 16 (per discussion here: https://discord.com/channels/140016142600241152/1320870710087913522).
- Declutters the last few bullet points of UDD (per my question here: https://discord.com/channels/140016142600241152/1320872790034743376).
- Adds a remark at the end of Lvl 16 for finding the focus of a 1 clue when it touches multiple unknown 1s.

### DIFF
--- a/docs/level-14.mdx
+++ b/docs/level-14.mdx
@@ -14,7 +14,7 @@ import TrashPush from "./level-14/trash-push.yml";
 
 - Sometimes, players will have two or more clued known-trash cards in their hand. (And everyone on the team can equally see that they are known-trash.)
 - In this situation, players are expected to discard the leftmost trash card.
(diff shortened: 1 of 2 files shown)


Zamiell:
a PR should only do one thing.
making a new small section of "Known-Trash Discard Order" seems fine.
however, i think the edits to level 16 are unrelated, so i removed them.
please do another pr if you want to edit that level.

PR #1300, opened 2025-01-23
Fixes nodejs vulnerable packages

springcomp:
This pull request ran the `npm audit fix` command and patched vulnerable packages.


Zamiell:
fyi you can pretty safely ignore those vulns because github does not use javascript to host the website.
and when you run the website on localhost it isn't accessible from the internet so it wouldn't be possible for anyone to hack you.

PR #1302, opened 2025-01-24
Challenge questions for level-6

springcomp:
This pullrequest adds a couple of challenge questions for level-6.
It attempts to give clear examples for:

- A valuable _Tempo Clue_ that gets two new cards to become playable.
- A non-valuable _Tempo Clue_ that _Chop Moves_ the card on the chop.

### DIFF
--- a/docs/challenge-questions/level-6-question-1.yml
+++ b/docs/challenge-questions/level-6-question-1.yml
@@ -1,0 +1,27 @@
+stacks:
+  - r: 4
+  - y: 2
+  - g: 2
+  - b: 2
+  - p: 5
+players:
+  - cards:
+      - type: x
+      - type: x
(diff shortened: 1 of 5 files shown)


Zamiell:
if you click on the red x you can drill down to the build error, which is this:
```
Exhaustive list of all broken links found:
  - Broken link on source page path = /challenge-questions/level-6:
     -> linking to beginner/minimum-clue-value-principle (resolved as: /challenge-questions/beginner/minimum-clue-value-principle)
```

Zamiell:
also btw, you don't have to force push, you can just keep pushing normal commits and they will be squashed

Zamiell:
nice work, i will accept this for now and perform some further edits later on

PR #1304, opened 2025-01-27
Reorder trash bluff and trash finesse

demipixel:
Currently, _Trash Bluff_ is taught before _Trash Finesse_. This order doesn't make sense for a variety of reasons. _Trash Bluff_ is simply a combination of _Trash Finesse_ and _Bluff_, so it should be taught last.

Changes:
- Move _Trash Finesse_ to be first
- Update _Trash Finesse_ example to have two missing 1s to clearly indicate that only *one* blind-play is needed to indicate all clued cards are trash.
- Move _Reverse Trash Finesse_ to follow _Trash Finesse_, since a regular _Reverse Finesse_ is level 2 (bluffs aren't until level 11)
- Update _Trash Bluff_ verbiage to fit with the new order.


Zamiell:
i originally put it in that order because trash bluff is the simpler thing, but i am also fine with this ordering, thanks.

PR #1306, opened 2025-01-29
Adding 'finesse-able' caveat to 5 stall FPE in flowchart

HungryBradbury:
As per discussion here - https://discord.com/channels/140016142600241152/1329097110771077174/1329128810628644976 I've added that the card on f pos needs to be finessable. Of course I didn't wanna fail CI test so changed 'finessable' to 'finesse-able'.


Zamiell:
i changed it back to finesseable for now

PR #1307, opened 2025-01-31
Fix minor errors in level 14

RandomGuyJCI:
Fixed incorrect formatting and naming (changed trash bluff to trash finesse)


Zamiell:
great pr, thank you!

PR #1309, opened 2025-02-05
Minor fix to an edit of lvl 7 flowchart

HungryBradbury:
An earlier commit by James split the starting node into two, this commit slots them into their respective classes to give them color.


Zamiell:
oops sorry

PR #1310, opened 2025-02-09
Reword out-of-order condition on tempo clues

3geek14:
With the old wording, I interpreted "out of order" to only apply to cards in the same suit, where one of them _must_ play before the other _can_ play. With multiple unknown 2s, there's no order in which the cards are required to play, but the left one can still block a prompt on the right one, preventing a playable 3 from being clued.

### DIFF
--- a/docs/level-6.mdx
+++ b/docs/level-6.mdx
@@ -20,7 +20,7 @@ import DiscardModulation from "./level-6/discard-modulation.yml";
 
 - In some special situations, _Tempo Clues_ have enough value to justify giving them:
   1. When everyone knows it gets two or more clued cards to play (in total across all hands).
-  1. When the clued card cannot be _Prompted_ (because there is another unplayable clued card in the hand that would "block" the _Prompt_).
+  1. When the clued card is not a 5 and cannot be _Prompted_ (because there is another unplayable clued card in the hand that would misplay if the _Prompt_ was attempted).
   1. When playing the clued card would "unlock" someone's hand. (For level 9 players, see the section on [_Locked Hands_](level-9.mdx#the-locked-hand-save-lhs).)
 - If a _Tempo Clue_ is given in any of these situations, it is considered to be "valuable".
 - _Valuable Tempo Clues_ can be given at any time. They don't have any special rules associated with them and they are treated in exactly the same way a "normal" _Play Clue_ is.
--- a/docs/level-6.mdx
+++ b/docs/level-6.mdx
@@ -20,7 +20,7 @@ import DiscardModulation from "./level-6/discard-modulation.yml";
 
 - In some special situations, _Tempo Clues_ have enough value to justify giving them:
   1. When everyone knows it gets two or more clued cards to play (in total across all hands).
-  1. When the clued card is not a 5 and it is "out of order" (meaning that it is not possible for a _Prompt_ to get the card played at this moment in time).
+  1. When the clued card can't be _Prompted_ because another card is blocking it (such as when g2 and g3 are "out of order" and both already clued with green).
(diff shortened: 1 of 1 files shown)


HungryBradbury:
How about 'When the clued card cannot be _Prompted_ because it is to the **right** of a similarly-clued card'? This way you won't need to include an example of color and rank.

HungryBradbury:
I like how your commit removes the 'not a 5' point, because it becomes self-evident.

3geek14:
> How about 'When the clued card cannot be _Prompted_ because it is to the **right** of a similarly-clued card'? This way you won't need to include an example of color and rank.

I like this wording for NoVar. I worry about variants like rainbow. Suppose two cards have positive red, and one of them also has negative green. Are those "similarly-clued"? The left one might block the right one.

Zamiell:
see my edits

3geek14:
> > > How about 'When the clued card cannot be _Prompted_ because it is to the **right** of a similarly-clued card'? This way you won't need to include an example of color and rank.
> > 
> > 
> > I like this wording for NoVar. I worry about variants like rainbow. Suppose two cards have positive red, and one of them also has negative green. Are those "similarly-clued"? The left one might block the right one.
> 
> In multi-var if both cards have red clues, and the left card (newer card) has neg green, then the right card (older card) must also have some new info as a consequence. For example, clearly the left card is a red card, and the right card could be revealed as a red card as well (if it received neg g info) or multi card (if it received the green clue).

The right card could have negative green from before the left card entered the hand, and a red prompt would get the left card, because we use "most positive information" for prompts.

3geek14:
> see my edits

You added that the left clued card isn't playable right now. That's asymmetric information, right? In rainbow, if I have m4 r3 in my hand, both clued with red, and both are playable, how do I interpret a 3 clue to me? Everyone else can see that m4 is playable, so this is not valuable. But for all I know, it's r4, and this is a valuable clue. Do I move my chop?

HungryBradbury:
> > > > How about 'When the clued card cannot be _Prompted_ because it is to the **right** of a similarly-clued card'? This way you won't need to include an example of color and rank.
> > > 
> > > 
> > > I like this wording for NoVar. I worry about variants like rainbow. Suppose two cards have positive red, and one of them also has negative green. Are those "similarly-clued"? The left one might block the right one.
> > 
> > 
> > In multi-var if both cards have red clues, and the left card (newer card) has neg green, then the right card (older card) must also have some new info as a consequence. For example, clearly the left card is a red card, and the right card could be revealed as a red card as well (if it received neg g info) or multi card (if it received the green clue).
> 
> The right card could have negative green from before the left card entered the hand, and a red prompt would get the left card, because we use "most positive information" for prompts.

That's true, but I don't see an issue here. Could you give me a hypo?

Zamiell:
> You added that the left clued card isn't playable right now. That's asymmetric information, right?

yes, it is (unfortunately) possible for TCCM to be asymmetric in this way.
that's just how the current convention is.

Zamiell:
also btw i did this commit too: https://github.com/hanabi/hanabi.github.io/commit/b237932f8cda5db149f9106215ab8f9d057eab26

3geek14:
> > You added that the left clued card isn't playable right now. That's asymmetric information, right?
> 
> yes, it is (unfortunately) possible for TCCM to be asymmetric in this way. that's just how the current convention is.

Gotcha. I'll move to Discord, because I think that's worth discussing and potentially changing.

> That's true, but I don't see an issue here. Could you give me a hypo?

I think your suggested wording of "similarly clued" requires that "positive red" and "positive red, negative green" be interpreted as "similar". I think that's a good interpretation, and it's probably the one most readers would have, but I wanted something more explicit, like what Zamiel went with.

I guess my "hypo" is: someone reading the docs misinterprets "similarly clued".

PR #1313, opened 2025-02-10
Reword 5cm "one away from chop"

3geek14:
Per discussion at https://discord.com/channels/140016142600241152/1330511255445962812, some beginners were misinterpreting "one away from chop", thinking it meant "the card directly left of chop (clued or unclued)". I don't love this new wording, but I think it's more clear.


Zamiell:
see my edits

PR #1314, opened 2025-02-11
Make zoom step on flowcharts configurable

maxcabrajac:
the default scroll_step was 0.2. I just set it to 0.5, but we can quickly change it if needed.


Zamiell:
thanks!

PR #1315, opened 2025-02-12
Updated 5’s Chop Move convention

Rewarsd:
With the change to 5 Stalls, players may now use them even when other clues are available all the way up to Level 9. Thus the block of text talking about analyzing if the clue-giver had another viable clue to figure out if a the 5 clue is a Stall or a Chop Move is relying on out-of-level rules about 5 Stalls. The 5’s Early Chop Move convention in Level 19 reintroduces the idea from scratch, so nothing is lost from deleting this paragraph.


Zamiell:
excellent, thank you!

PR #1316, opened 2025-02-14
Unify Pink Promise with 5 Stall/5 Chop Move and move them up

kernel18:
It seems weird to me that "Pink Promise with 5 Stall" and "Pink Promise with a 5 Chop Move" are listed under Edge Cases (classified as level 21). Because these conventions say essentially the same thing -- that's why I unified them under a single section -- and they relate to common clues, I think they should be moved up under "Basic pink principles" so that people starting to play Pink are aware of them. This PR also include minor edits to "Pink Promise with a 5 Pull" (added a comment on the rare 3-player case of a 5 Pull performed on slots 1 and 3.)

### DIFF
--- a/docs/variant-specific/pink.mdx
+++ b/docs/variant-specific/pink.mdx
@@ -401,7 +401,7 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 
 <FakeFinesse />
 
-## Edge Cases
+## Level 21 - Edge Cases
 
 ### Trash Moves in Pink
 


docs/variant-specific/pink.mdx
```diff
@@ -365,7 +401,7 @@ These conventions apply to any variant with a pink (touched by all ranks) suit.
 
 <FakeFinesse />
 
-## Level 21 - Edge Cases
```
Zamiell, line 368: why did you make this change

Zamiell:
i'll see if i can take a look at this one this weekend

Zamiell:
sorry for taking so long to look at this.
the philosophical idea of having an "edge cases" is so that people can read LESS and jump into the action.
your PR goes against this goal, so thus I think we should probably close it for now.

> This PR also include minor edits to "Pink Promise with a 5 Pull" (added a comment on the rare 3-player case of a 5 Pull performed on slots 1 and 3.)

the new reminder you added doesn't hurt.
but on the other hand it isn't strictly necessary since it is just a repeat of information that is already contained in the main 5 Pull section.
i generally like to reduce the amount of duplication in the doc, since it makes it hard to keep in sync when things are updated.
thus I don't think we should integrate this change either.

PR #1318, opened 2025-02-17
Level 1 - 2 Save - fixed typo.

springcomp:
This pull request fixes a slight typo in Level 2, in the paragraph related to the _2 Save_ special move.

```patch
  By definition, you can only perform a _2 Save_ with a number clue. 
- (If a color clue was used to touch an unclued 5, then it would be a _Play Clue_ instead.)
+ (If a color clue was used to touch an unclued 2, then it would be a _Play Clue_ instead.)

```


Zamiell:
thx

PR #1319, opened 2025-02-19
Disable hotkey whenever an `<input>` is focused

TRCYX:
The website contains hotkeys such as `l` for jumping across levels. Currently, they are only disabled when the search panel pops up. There is also a `/search` page, and the user can enter that page from the search panel when there are too many search results (for example when the search string is a single letter). Under `/search`, the user can further modify the search string in a larger `<input>`. However, hotkeys are not disabled for this page, which makes typing `l` there quite problematic.

This pull request instead disables hotkeys whenever an `<input>` is focused, since all current hotkeys contribute to text typing. There does not seem to be `<textarea>`'s or other editable elements, so this should suffice.

### DIFF
--- a/static/js/hotkey.js
+++ b/static/js/hotkey.js
@@ -18,8 +18,8 @@ function main() {
       return;
     }
 
-    // Do not do anything if we have the search box focused.
-    if (isSearchBarFocused()) {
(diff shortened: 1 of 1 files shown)


Zamiell:
thanks

PR #1321, opened 2025-03-05
level 4: chop move focus exception

springcomp:
This pull request updates the game state diagram for the Level 4 section titled **[Chop Moves & New Clues](https://hanabi.github.io/level-4#chop-moves--new-clues)** related to clarifying the focus of a clue that touches a _Chop Moved_ card for the first time.

image

It specifically highlights that the two previously unknown "red" cards are retouched and, thus, the clue only touches old cards and the _Chop Moved_ card for the first time.


Zamiell:
thx

PR #1322, opened 2025-03-06
Fix image-text mismatch in level 14

TRCYX:
I notice that level 14 contains a few mismatch between text and image in examples. I choose to change everything in the text to match the images.

Also, I feel there's a grammatical mistake on line 124 (an extra "with"), so I modify the sentence to make it more fluent.


Zamiell:
thx

PR #1323, opened 2025-03-15
Clarifying precedence from ambiguous TCCM over TCM

springcomp:
This pull request adds a [clarifying principle](https://discord.com/channels/140016142600241152/1350409638478217237) in level-6 to defined precedence of _Tempo Clue Chop Moves_ over _Trash Chop Moves_ when interpreting an ambiguous clue.

This pull request adds the following section:

### The Ambiguous Tempo Clue Chop Move

- _Trash_ cards may interact with _Tempo Clues_ in a potentially confusing way.
- Consider the following situation:

  - The red 1 card is played on the stacks.
  - Alice has a red card in slot 3, known to be a playable red 4.
  - Bob has an unknown 5, saved on chop.
  - Cathy has a red card in slot 2, known to a a playable red 3.
  - Alice then gives a red clue to Bob, that touches a red card on slot 1 and fills-in the previously unknown red 5 in slot 5.

image

- In this situation, the focus of the red clue is ambiguous:

  - If the red clue were focused on the "new" slot 1 card, then the focus would be _Trash_, so the resulting move would be a _Trash Chop Move_ on slots 2, 3 and 4.
  - If the red clue were focused on the already-clued slot 5, then the focus would be the previously-unknown-to-be-playable red 5, so the resulting move would be a _Tempo Clue Chop Move_ on slot 4.
  - We agree that _Tempo Clue Chop Moves_ have precedence over _Trash Chop Moves_.
  - Thus, in this situation, Bob should only _Chop Move_ the card in slot 4.
  - This follows from our general approach of "when interpreting an ambiguous clue, use the most conservative interpretation possible".


Rewarsd:
Bob’s slot 4 is marked as chop moved in the diagram but not stated to be chop moved in the text. Perhaps change either to make the text and diagram consistent?

springcomp:
> Bob’s slot 4 is marked as chop moved in the diagram but not stated to be chop moved in the text. Perhaps change either to make the text and diagram consistent?

I think this is consistent with the various examples about _Chop Moves_ in the documentation.

For instance, [the first example](https://hanabi.github.io/level-6#the-tempo-clue-chop-move-tccm) in level-6 has the following diagram and corresponding text:

image

_"… Since it does not get any new cards and only get Tempo on one card in total, he knows that he should also Chop Move his slot 5 card."_


In this pull request, the accompanying text states:

_"… Bob should only Chop Move one card."_

Do you think changing this to:

```patch
- … Bob should only Chop Move one card.
+ … Bob should only Chop Move the card in slot 4.
```

Would be better?
What do you suggest I should change?

Rewarsd:
I think this change makes it clearer what the consequence of the clue should be, awesome

Blaizer:
Couple mistakes I noticed:

> so the resulting move would be a Trash Chop Move on slots 2 and 3.
This should say slots 2, 3 and 4.

> so the resulting move would be a Tempo Clue Chop Move on slot 3.
This should say slot 4.

HungryBradbury:
The last line says that TCCM is more conservative than TCM, I don't agree with that. It's safer to chop-move more cards than fewer. I think the line should say something along the lines of 'filling up a person's hand increases their risk of getting locked in the future, which is discouraged', in your own words ofc

Zamiell:
>  I don't agree with that. It's safer to chop-move more cards than fewer.

Well, to be fair, when I wrote the words "use the most conservative interpretation possible", I did NOT really mean "use the safest interpretation". What I really meant was: "use the interpretation that does the smallest thing or least complex thing".

Specifically, _Tempo Clues_ are just a form of a normal _Play Clue_. Players perform _Tempo Clues_ all the time, even naturally on their very first games of Hanabi without knowing any prior conventions! In contrast to that, _Trash Chop Moves_ are an arbitrary invention that we made that has specific mechanics and is "unnatural". Thus, I think _Tempo Clues_ count as the least complex thing.

I'm not sure how the best way to encapsulate this sentiment into the PR, but hopefully that clarifies at least what I originally meant.

HungryBradbury:
> Players perform _Tempo Clues_ all the time, even naturally on their very first games of Hanabi without knowing any prior conventions! In contrast to that, _Trash Chop Moves_ are an arbitrary invention that we made that has specific mechanics and is "unnatural". Thus, I think _Tempo Clues_ count as the least complex thing.

This makes sense, thanks!

Zamiell:
thinking about this PR more, i would prefer not to add a ton of text to the doc just to handle this special case.
(an overarching goal of the docs is to keep it as short as possible.)
how do you feel about adding one line to the TCM section that simply says "Tempo Clues take precedence over Trash Chop Moves".

Zamiell:
> Do I move that to a challenge question in level-6 ?

yes please

Zamiell:
great work thank you!

PR #1324, opened 2025-03-21
Lvl 17 YML fixing Retouched status

HungryBradbury:
In TTCM Blind-Play Form example, the clue to r3 is a reclue, but the YML diagram shows it as a new clue.


Zamiell:
thx

PR #1325, opened 2025-03-21
Level 4 Challenge Question: Fresh 1's and OCM

springcomp:
This pull request suggests a Challenge Question for Level-4.

It illustrates a situation where, even though 1's are played from right to left at the start of the game, playing "fresh" 1's should take precedence over playing 1's from the starting hand. As a result, this performs an _Order Chop Move (OCM)_ on the very next player.

image

### DIFF
--- a/docs/challenge-questions/level-4-question-1-part-2.yml
+++ b/docs/challenge-questions/level-4-question-1-part-2.yml
@@ -41,14 +41,14 @@ players:
   - name: Bob
     cards:
       - type: x
-      - type: 1
-        clue: 1
+      - type: x
+        middleNote: (1)
         below:
           text:
             - Fresh
           color: Green
-      - type: 1
-        clue: 1
+      - type: x
+        middleNote: (1)
         below:
           text:
             - Starting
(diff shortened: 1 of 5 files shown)


HungryBradbury:
Thanks for taking the lead on this one springcomp. 
I'm not sure if I get this example (the one in the PR description, have yet to take a look at your commit). You can refer to my PR's 1 Clue Stacking example

HungryBradbury:
Oh I see, this PR isn't for 1 Clue Stacking. I'll whip something up over the weekend then =D

springcomp:
@HungryBradbury  that’s right, this is an example for the fresh 1’s rule.
Specifically, how it relates to OCM beinf performed.

Zamiell:
actually the question is bugged, because Alice needs to be the person who takes the first action, and right now it is Cathy.
so you need to move Alice to Bob and Bob to Cathy and Cathy to Bob.

springcomp:
@Zamiell  I have updated the PR accodingly.

Zamiell:
taking a look now

Zamiell:
thanks again and sorry for taking so long

PR #1327, opened 2025-03-29
Add ambiguous GD to level 10

3geek14:
If Bob doesn't fully know the identity of his card, but all possible options are on other players' _Finesse Positions_, he may initiate a _Gentleman's Discard_ by discarding his card. This results in a single player blind playing their _Finesse Position_.


hm1298:
Copying over from Discord. This is probably solved with a smaller change, since I think the example already does a good job of explaining what is illegal about the situation; just add this (or similar) beneath the original Gentleman's Discard definition & picture:

```- In rare cases, Cathy may perform a _Gentleman's Discard_ on a card that appears to have multiple possible identities. Only the true identity is signaled to play.```

Solves the issue caused by outlining a legal clue under a section entitled Common Mistakes, as 3geek13 brought up. I don't think I can edit directly without opening a new PR myself

Zamiell:
thanks

PR #1330, opened 2025-05-12
Level 14 convention reordering

Rewarsd:
The level 14 text goes Trash Finesse -> Trash Bluff -> Trash Push. Rather confusingly, the Trash Finesse section mentions both others, despite coming before them. Reordering them to go Trash Push -> Trash Finesse -> Trash Bluff removes one inconsistency, while the other one is simply deleted. I've also switched around the order of the Forced Gentleman's Discard Chop Move and Trash Bluff sections, simply to put the former next to the Reverse Trash Finesse one.


Zamiell:
lgtm thanks

PR #1331, opened 2025-05-12
Update convention-accumulation.md

vEnhance:
Think a couple stray asterisks got added in from back when this was bold


Zamiell:
thx

PR #1332, opened 2025-05-20
Fixing contradictory statement between Brown's Tempo Clue Inversion & Mud Clues

HungryBradbury:
The section Tempo Clues in Muddy/Cocoa Rainbow variant specific convention doc makes a false statement that when two or more muddy/cocoa rainbow cards are given a tempo clue, then the rightmost one plays. This is Brown's Tempo Clue Inversion, and is contradictory with Mud Clues.


Zamiell:
lgtm thanks

PR #1333, opened 2025-05-20
Replacing Hard Color Bluff with CCP in Prism

HungryBradbury:
Per discussion here (https://discord.com/channels/140016142600241152/1373755373458165930) it was felt that simply formalizing Cathy's Connecting Principle in Prism is enough to make a Hard 3 Bluff of prism 1 through a color clue possible, and so there is no need to make a new convention for it.


Zamiell:
seems fine thanks

PR #1340, opened 2025-07-04
Update Pink Fix Clues

3geek14:
In the examples, Bob plays 1s that were in slots 2 and 3 at the time of being clued. When Bob plays the first 1, the second 1 moves into slot 3. So Bob plays from slot 3 twice in a row, rather than from slot 3 and then from slot 2.


Zamiell:
thx

PR #1342, opened 2025-07-30
Update assuming-asymmetric-information.yml

Gingy-120:
Alice has one too many cards.


Zamiell:
thank you, i missed this earlier because i am bad

PR #1347, opened 2025-08-26
Fixes order of cards in Bob’s hand.

lanijw:
The text says "Cards are drawn and added to the left side to each hand. Thus, Alice's newest card is the blue 1 and Bob's newest card is the red 1.” I believe this means that Bob’s red 1 should be in slot 1 instead of 5.  
Live view: https://hanabi.github.io/beginner/card-layout


Zamiell:
actually lets just update the text instead of the diagram.

Zamiell:
i'll do it

PR #1356, opened 2025-09-09
Clarify Shift Focus Answer

d20cay:


3geek14:
As the question currently says, if s1 is focused, then this is a tcm. The question also claims the focus is ambiguous, which I disagree with. The algorithm for determining focus is pretty straightforward, which is why (in another recent discussion), we see that trash finesses and trash chop moves can cm different numbers of cards in the hand: we don't "re-evaluate the focus" after seeing the finesse. In order to claim that s5 is the focus in the question, I think we need tccm vs tcm preference to be an example of focus shifting.

d20cay:
@3geek14 I haven’t learned anything past level 6 yet, so I didn’t understand a lot of what you said, but that’s totally ok. I totally understand wanting to discuss at a higher level.

@Zamiell I Does that mean the solution would be that slot 2,3, and 4 would be chop moved with a tcm?

PR #1386, opened 2025-12-18
Move images to subfolders

bnordli:
All images in the "basic", "advanced" and "variant" sections are put into separate subfolders, based on the page they're on.

Currently, the images are really cluttering up the "beginner" and "challenge-questions" folder, so I think it is useful to create subfolders with images there too.


bnordli:
It is reasonable, but I think it will be a bit cumbersome to maintain that structure.

But it's ultimately your call. I can modify this PR if you want to have it that way.

Zamiell:
ill just accept the PR as-is then

PR #1388, opened 2025-12-19
More image enhancements

bnordli:
More enhancements to the image generation script:

1. Card option to draw wrench/fix icon. (Image pulled from https://github.com/Hanabi-Live/hanabi-live/blob/main/public/img/wrench.png.)
1. Card option to draw chop move border.
1. For "above" and "below" attributes, support plain string array in addition to "text:" attribute for multiline text.
1. Expand some common shortcuts to full text.
1. More color definitions.

Also document the new features.

Added suggested guidelines for example images. (Feedback is appreciated.)

### DIFF
--- a/docs/example-images/card-text.yml
+++ b/docs/example-images/card-text.yml
@@ -18,7 +18,9 @@ players:
         above: Red 1
       - type: x
         above:
-          text: [Multi-, line]
+          text:
+            - Multi-
+            - line
         below: Focus
       - type: x
         above: Rainbow 4
@@ -26,7 +28,9 @@ players:
   - cards:
       - type: x
         above:
-          text: [Foo, Bar]
+          text:
+            - Foo
+            - Bar
           color: rainbow
       - type: x
         above:


Zamiell:
The changes to `above` and `below` are confusing. Why does the type include both a string array and a normal array? I think it is more idiomatic to use a normal YAML array and not a string array. Specifically, I am talking about this:

```
   above: z
      .string()
      .min(1)
      .or(z.array(z.string().min(1)).min(1))
      .or(textObject)
      .optional(),
 ```
 
 It should probably just support one or the other, and a YAML array might be preferrable over a special string that gets cast to a TypeScript string array?

Zamiell:
> I thought this syntax was a bit cumbersome to write
> 
> // syntax 1
> above:
>   text:
>     - Line 1
>     - Line 2
> ..and wanted to also support this syntax:
> 
> // syntax 2a
> above:
>   - Line 1
>   - Line 2
> ..which is equivalent to:
> 
> // syntax 2b
> above: [Line 1, Line 2]

There is a tradeoff though. Do the pros outweigh the cons? The pros are that, when you don't need to specify colors, you can be slightly more terse in your line representation, which saves some lines of code and is more convenient. But the con is twofold. First, we increase the complexity of the Zod types, which becomes more difficult to maintain in the long term. Second, by supporting all of these different formats, it becomes unclear which format you should use when writing new code.

In this case, I probably lean towards reducing type complexity rather than getting a marginal benefit in convenience. Meaning that we should probably keep the existing 2 formats and not add any more.

The rest of the PR looks fine though.

PR #1401, opened 2025-12-21
Image enhancements, Level 3

bnordli:
Added/fixed card notes and explanations below cards.
Notes:
- Added a new example for the second sarcastic discard
- playing-multiple-1s-3: Swapped players (Alice should be the clue giver by default)
- Removed unnecessary comments about variants and number of players.


Gingy-120:
So about this section:
https://hanabi.github.io/level-3/#playing-multiple-1s

The text says Alice tells Bob about three 1s but the image shows Alice having the three 1s. What is the intended layout?

Zamiell:
you can't nest PRs, so you just have to create a new one

PR #1422, opened 2025-12-23
Link from level 14 to correct section in level 24

bnordli:


bnordli:
I don't play on level 24, so I was very confused by this link. After rereading and checking, I am quite sure _this_ is the extra thing that happens which should be referred to.

(I am not really sure why a forward link from level 14 to level 24 is necessary in the first place, though..)

bnordli:
It says "an extra thing happens".

If the trash card is _not_ on chop. Level 14 chop moves all cards to the right. This doesn't change on Level 24.
https://hanabi.github.io/level-24#unnecessary-moves-with-unknown-trash-off-chop----chop-move says:
> As explained in the first section, Trash Finesses and Trash Bluffs are always unnecessary, so this comment does not apply to them. (But it applies to the other fancy trash moves.)

So I believe this bullet point _only_ applies if the clued card is _already_ on chop. What happens then is explained in https://hanabi.github.io/level-24#unnecessary-moves-with-unknown-trash-on-chop----trash-push

The link goes to https://hanabi.github.io/level-24#trash-finesses-and-trash-bluffs-are-always-unnecessary, which (in practice) only says

> However, we explicitly agree that Trash Finesses and Trash Bluffs are always treated as being unnecessary, regardless of whether it was actually possible to "get" the blind-played card with a normal clue.

..which is the same as is said on level 14:

> After a Trash Finesse, the player who was clued should Chop Move all of the cards to the right of the trash card. (In other words, you can think of a Trash Finesse as both a Finesse and a Trash Chop Move at the same time.)

-- 

If this is incorrect, you will have to explain it to me :)

PR #1423, opened 2025-12-23
Image enhancements, Level 24

bnordli:
Notes: All trash finesses are now unnecessary, so some unnecessary words are removed.
(I am also not sure whether this example is needed now; it's basically a reiteration of a level 14 move.)


Zamiell:
i agree that reiterating the exact same example from level 14 is of questionable value.
but maybe it is worth it, just as a review for the reader, and also serves as an introduction to the next sections.
/shrug

PR #1447, opened 2025-12-26
swap Alice and Bob due to bug

Gingy-120:
swap Alice and Bob due to bug


bnordli:
Sorry, thanks for clearing this up! :)

Issue #1451, opened 2026-01-02
Expand Glossary


bnordli:
The [Glossary](https://hanabi.github.io/glossary) exists, but is pretty bare bones.

I suggest to add some of the following definitions to the list:

* Chop
* Focus
* Card Notes
* Chop Move
* Clues
  - Save
  - Play
  - Stall
  - Fill-In
  - Burn
  - Trash
  - Chop Move
  - Tempo
  - Finesse
  - Prompt
  - Bluff
  - Push
  - Pull
  - Unnecessary
* Good Touch Principle
* Save Principle
* Minimum Clue Value Principle
* Locked Hand
* Positions
  - Finesse (already covered)
  - Ejection
  - Discharge
  - Charm
  - Ignition
* Phases
  - Early Game
  - Low Score Phase
  - End-Game
* Context
* Elimination
* States
  - Occupied
  - Loaded
  - Double Discard
  - Locked
  - Anxiety
  - Stalling

A decision must be made whether to keep the list alphabetically sorted (like a real glossary), or broken into clues/phases/states outlined here, or by level/complexity.

Yes -- I know there is a search feature, but it doesn't really work for basic concepts. For instance: The first result for "Finesse" goes to "Examples of the Surreptious Finesse"..

PR #1452, opened 2026-01-02
Expand the landing page with more links into the documentation

bnordli:
* Add more links to the top of the landing page (and each page)
* Link to the glossary on the main page, making a 2x2 grid instead of a 1x3 grid.

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -46,6 +46,12 @@ const config: Config = {
       },
       items: [
         {
+          to: "about",
+          activeBasePath: "docs",
+          label: "About",
+          position: "left",
+        },
+        {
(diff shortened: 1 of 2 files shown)


Zamiell:
Do other Docusaurus websites have more than 3 landing items?
I think we stylistically want to keep the landing page to 3 items to make it cleaner.

bnordli:
I don't know of any existing guidelines or best practices for Docusaurus. And I can't easily find any examples either.

But if you think 3 items is enough, not adding the link to the Glossary is fine with me. Or we can in any case wait until it is has a bit more content (re https://github.com/hanabi/hanabi.github.io/issues/1451).

PR #1464, opened 2026-01-04
Challenge level 5

### LINKED ISSUE #1404: More ambiguous examples for Level 5
There's the https://hanabi.github.io/level-5#the-ambiguous-finesse section, but IMO ambiguity is a common source of confusion, so I will suggest to add at least

* an example (or challenge question) on how to _avoid_ ambiguity (with a number clue instead of a color clue and/or vice versa)
* an example (or challenge question) that highlights what should happen if a seemingly finessed player discards instead of blind-playing

I don't think this level should cover "suboptimal" actions other than the _Urgency Principle_ for saves, so in these examples it should be _pretty_ clear that the finesse was "missed".

bnordli:
Closes #1404.

New challenge questions for level 5.
(Please check my reasoning and text.)
- What's Hidden: Demonstrating the Hidden Finesse
- The Onion: Demonstrating a mixture of Hidden, Double, Clandestine and Layered finesses
- Finesse after Finesse...: Demonstrating a Queued Finesse
- A Tricky Spot, I: Demonstrating how to avoid ambiguity
- A Tricky Spot, II: Demonstrating that ambiguity is sometimes necessary
- Bob is Lazy: Urgency principle and how to react after an ambiguous clue

### DIFF
--- a/docs/challenge-questions/level-5-what-s-hidden.mdx
+++ b/docs/challenge-questions/level-5-what-s-hidden.mdx
@@ -22,22 +22,21 @@ import Answer from "./level-5-what-s-hidden/answer.yml";
 - The red 1 and blue 1 are played on the stacks.
 - Bob has:
   - An unclued blue 2 on slot 1.
-  - A red 2 on slot 4 clued with a number 2 clue.
-- Alice clues number 3 to Cathy, touching a 3 as a _Play Clue_.
-- Bob plays the blue 2.
-- Which notes should Cathy write?
-- What should she do next?
+  - A red 2 on slot 4 clued with a number 2 clue. (Bob does not know the identity.)
+- Alice clues number 3 to Cathy, touching a 3 on slot 3 as a _Play Clue_.
+- Bob plays the red 2.
+- Which notes should Cathy write on her 3?
+- What should Cathy do next?
 
 <Question />
 
 </TabItem>
 <TabItem value="solution">
 
-- Cathy's card might be a red 3, in which case this is just a normal [_Prompt_](../level-1.mdx#the-prompt) on the red 2.
-- Her card might also be a blue 3, if Alice intended a [_Hidden Finesse_](../level-5.mdx#the-hidden-finesse) on the blue 2.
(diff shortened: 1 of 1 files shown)


Zamiell:
also see this commit: 369d2d1e7d0299ae0ed1849ef9493eff585a19ec

bnordli:
I'd rather rename the challenge questions for the _other_ levels to match this: I think it's more interesting to have a question title than just "Question N", especially since there now is a progress bar.

(Level 3 and 4 already have question titles.)

Also, if someone creates an interesting new question and wants to insert it into the middle of the pack, reordering the others will be a chore. (Of course, it can be scripted.)

I will take a closer look of the styling.

Zamiell:
> I'd rather rename the challenge questions for the other levels to match this: 

eh, i thought about that, but
1) its more work
2) i don't really like how the title of the question is a hint for what the answer is.

for an example of what I mean, check out this question:
https://hanabi.github.io/beginner/5-save-question-1?tab=question

This one is called - "The 5 Save (Question 1)"
But the answer is that you shouldn't 5 save.
If the question was cheekily called "The 5 Save - Doesn't Apply Always" or "The 5 Save - You Don't Always Need to Clue 5" or something like that, I think it really detracts from the challenge of the question and makes it a substantially worse.

Zamiell:
> I think Beginner's Challenge questions are fine as they are. They are in the middle of other pages and need to stand out.
> But, I also think that a good challenge question writer should take the challenge (;)) and find a good title without spoiling (too much) about the question itself

fine

> What do you think about the current titles to questions on level 3 and 4?

They are fine i guess, I don't have the time to look at the questions too closely

> Renaming the other 16 questions is a one-time job and I volunteer to take a stab

ok

Zamiell:
also, "The Onion" implies a layered finesse, should we just call it "Multiple Possibilities" instead?

Zamiell:
> If that is too much for one question, I can try to reword it into only Cathy's perspective, or I can delete it, it's not the most important question here.

yeah i would prefer to have two different questions if we want to have two different perspectives.
we can try reword it to just cathy's perspective then? will the question still be substantial enough to be worth it?

PR #1467, opened 2026-01-05
Editing guidelines for the H-Group conventions 

bnordli:
Suggested guidelines for doc and convention writers.


Zamiell:
thanks, i made some edits

PR #1476, opened 2026-01-07
Challenge questions for level 7

bnordli:
- "Alice is Forgetful": Scream Discard Chop Move
- "Bob is Forgetful": The above has a Finesse Exception
- "Screaming into the Void": When _not_ to scream
- "The Silent Scream": Should Discard Chop Move
- "Alice loves Garbage": Scream Discard with Known Trash
- "Too Few Clues": Generation Discard
- "Another Scream Discard": Do _not_ scream for non-critical cards
- "Should this Blind-Play?": Blind-playing CM'd cards


Zamiell:
thanks for the PR, it is nice to have more challenge questions.

also see my edits.

PR #1477, opened 2026-01-13
Update brown.mdx

HungryBradbury:
It's a minor change to clarify that this applies on Tempo Clue Inversion and Double Tempo Clue Inversion also. The way this section is laid out, it seems Brown Tempo Clue is a separate topic from Tempo Clue Inversion and Double Tempo Clue Inversion, which doesn't make much sense.

### DIFF
--- a/docs/variant-specific/brown.mdx
+++ b/docs/variant-specific/brown.mdx
@@ -21,7 +21,7 @@ These conventions apply to any variant with a brown (touched by no ranks) suit.
 
 - Normally, if a _Tempo Clue_ retouches two or more cards, only the leftmost card should play.
 - When a brown color clue retouches two or more cards, only the rightmost card should play.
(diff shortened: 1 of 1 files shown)


HungryBradbury:
So this is why the placement of this line here is bad. This line clarifies that if a Finesse is attempted by recluing two brown cards, it still focuses on the rightmost brown card even though finesses are not tempo-clues. We can move this line to the next two sections and remove it from here also. I just proposed a lighter solution.

EDIT: To be clear, this is not my interpretation, this is how other players interpreted this line in one of the brown games I played recently, so I thought this convention should be made clearer (if it exists at all)

Zamiell:
ok, see my latest edits, do you agree with this PR now?

HungryBradbury:
Great

PR #1480, opened 2026-01-14
Changing Orange Double Finesse example to avoid confusion with Defiant Bluff

HungryBradbury:
This example changes when Orange is played with Defiant Bluff turned on. By changing this example to a reverse double finesse it works at all levels.


Zamiell:
we are actually deleting Defiant Bluff, which should make this PR unnecessary.

Zamiell:
(see the discussion in the orange thread on Discord)

PR #1486, opened 2026-01-31
Add Orange layered moves. Avoid orange confusion.

bnordli:
* Use "drag to discard" and "drag to stacks" for (believed) orange cards to avoid confusion
* Add _Orange Layered Finesse_ and _Orange Layered Gentleman's Discard_
* Clarify that an _Orange Baton Discard_ target is chop moved
* Add another example for _Orange Trash Discharge_
* Some small language cleanup


Zamiell:
thanks. see my edits though

PR #1489, opened 2026-02-05
Fix missing player offset

Gingy-120:


Zamiell:
> I'd prefer if Bob was playing a blue 1 instead of a red 1, and that way we don't have to draw the wrench on Cathy's red 1. Is there a reason why the red 1 is currently duplicated in the example?

make it the way you prefer, thanks : )

Gingy-120:
The content doesn't need to be updated (from what I can see) but I will do some grammar fixing.

PR #1491, opened 2026-02-14
Update critical-save-question-3.mdx

djmclaugh:
Add why the card could not be a yellow 5.
I forgot that a 5 save HAD to be done with a 5 clue and it took me a while to figure out why the 5 option was not mentioned.


Zamiell:
see my edits

PR #1500, opened 2026-06-24
fix: save principle example b2 notes

vEnhance:
blue cards are missing some possibilities

see recent discussion on Discord #general

### DIFF
--- a/docs/beginner/save-principle.mdx
+++ b/docs/beginner/save-principle.mdx
@@ -27,7 +27,7 @@ import SavePrinciple from "./save-principle/save-principle.yml";
   - If Alice clues number 3, it will violate _Good Touch Principle_ because it will touch the red 3 (and the red 3 is already played on the stacks).
   - Here, Alice knows that it is better to violate _Good Touch Principle_ than to lose the chance at a perfect score. Thus, Alice must choose between the lesser of two evils and decide between cluing blue and cluing number 3.
   - Alice decides to clue blue to Bob instead of number 3 because at least one of the blue 4's will be useful later on.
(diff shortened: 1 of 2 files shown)


EhDrone:
That isn't the solution to the example. This is in Save Principle section, if Blue 2 were played then there is no save clue - this is a play clue, and the other cards are noted for the remaining blue cards.

Either the example needs to be reworked, or the all of the cards need negative 2 on them?

vEnhance:
ah crap youre right. i'm not thinking.

vEnhance:
i guess we can also just edit the notes instead to accommodate b2 possibility? that's a deeper edit so i need to get back to my laptop later for that

vEnhance:
still need to update the main text. once i have my desktop again.

Zamiell:
ok, I will just let this one sit then for now, thanks

vEnhance:
added some text, hope it's not too wordy? let me know what you think @Zamiell 

(also, i accidentally pushed to the main branch at first, sorry about that. i force-pushed it back to where it was before)

do we prefer `b2, b3, b4, b5` to `b2345`? if so i can change that too.

Zamiell:
we prefer `b2, b3, b4, b5`, as it is less cryptic

vEnhance:
ok changed

vEnhance:
ah, the middle note "b2, b3, b4, b5" is too long here. what's the right way to handle that?

Zamiell:
not really sure TBH, i guess we haven't run into that before, i guess use `b2345` for now, even though it won't have symmetry with any of the other notes.

Zamiell:
thanks evan, this was a good fix

PR #1503, opened 2026-07-06
Add card note to critical save answer 2

riking:
Felt like the correct card note here could use some visual emphasis.


Zamiell:
thanks

PR #1504, opened 2026-07-06
Remove extra clue arrow from save-principle-answer-1.yml

riking:
This old clue arrow is distracting from the clue that Bob actually just gave


Zamiell:
thanks

PR #1505, opened 2026-07-18
Update level-3.mdx

EhDrone:
Fixing Alice and Bob mix-up in example for chop focus exception


Zamiell:
i already changed the picture so that it matches the existing text in 6b2b4282dfbcaee323270e18b81a3690f29b7ede

EhDrone:
That's fine too - its probably similar enough to other scenarios where Bob is the clue giver though not exactly the same. Ill close this as pic and text are in sync

PR #1506, opened 2026-07-19
Update "Give Play Clues over Save Clues"

riking:
Say Immediate Play Clue, because a delayed play clue may still result in the chop being discarded


Zamiell:
thanks

PR #1512, opened 2026-08-23
Add Bluff clue possibility to The Finesse (Question 1)

michaelbrundage:
Cluing the green 4 also works as a bluff to get the blue 1 played and communicate to Cathy that her card is the green 4.


Zamiell:
bluffs are not a level 1 convention, so you aren't allowed to mention it here, right?

David707x:
> bluffs are not a level 1 convention, so you aren't allowed to mention it here, right?

Agree, we should not be mentioning bluffs in the beginner guide, it will only lead to unnecessary confusion.

michaelbrundage:
I see. ok.

PR #1513, opened 2026-08-28
Simplify the process for determining clue focus

byorgey:
When I was first learning level 1 I found the description of determining clue focus quite confusing.  Saying it was a "4-step process" and talking about whether there were multiple new cards touched or not seemed unnecessarily complicated, but it took me a while to untangle and make sure I was understanding it properly.  The flowchart that's already there in the level 1 document is actually perfect --- this PR just changes the textual description of the process to match, in both level 1 and the beginner's guide (with an extra warning in the beginner's guide that there could be other new cards touched in addition to the chop).


Zamiell:
seems fine with me, thanks

PR #1516, opened 2026-09-03
Update special-fives.mdx

J-Fricke:
* color promise not required while still encouraged
* color play clue lie implicitly allowed now
* color Promise Finesse turned on same level as other suboptimal finesse/bluffs are enabled in the main convention


Zamiell:
see my edits

